PR #8, opened 2023-12-05
add import languages

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/commands/importExportEntities/entities/languages.ts
+++ b/src/commands/importExportEntities/entities/languages.ts
@@ -1,11 +1,150 @@
-import { LanguageContracts } from "@kontent-ai/management-sdk";
+import { LanguageContracts, LanguageModels, ManagementClient } from "@kontent-ai/management-sdk";
 
+import { serially } from "../../../utils/requests.js";
+import { notNull } from "../../../utils/typeguards.js";
 import { EntityDefinition } from "../entityDefinition.js";
 
-export const languagesExportEntity: EntityDefinition<ReadonlyArray<LanguageContracts.ILanguageModelContract>> = {
+const defaultLanguageId = '00000000-0000-0000-0000-000000000000';
+
+export const languagesEntity: EntityDefinition<ReadonlyArray<LanguageContracts.ILanguageModelContract>> = {
   name: "languages",
   fetchEntities: client => client.listLanguages().toAllPromise().then(res => res.data.items.map(l => l._raw)),
   serializeEntities: collections => JSON.stringify(collections),
-  importEntities: () => { throw new Error("Not supported yet.")},
-  deserializeEntities: () => { throw new Error("Not supported yet.")},
+  deserializeEntities: serialized => JSON.parse(serialized),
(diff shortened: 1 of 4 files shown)


src/commands/importExportEntities/entities/languages.ts
```diff
@@ -1,11 +1,140 @@
...
+      })
+      .toPromise();
+  }
+}
+
+
+const updateFallbackLanguages = async (client: ManagementClient, importLanguages: ReadonlyArray<LanguageContracts.ILanguageModelContract>) => {
+  const projectLanguages = (await client.listLanguages().toAllPromise().then(res => res.data.items));
+
+  for (const lang of importLanguages.filter(l => l.id !== defaultLanguageId)) {
```
JiriLojda, line 121: What would you think about a structure like this instead?
```ts
await serially(importLanguages
  .filter(isNotDefault)
  .map(lang => {
    const projectLanguage = ...;
    if (!projectLanguage) {
      throw new Error(...);
    }
    return [lang, projectLanguage] as const;
  })
  .map(([lang, projectLanguage]) => async () => {
    const projectFallbackLanguage = projectLanguages.find(l => l.id === projectLanguage.fallbackLanguage?.id);
    const importFallbackLanguage = importLanguages.find(l => l.id === lang.fallback_language?.id);
    
    if (...) {
      await ...
    }
  }));
```
It separates the different actions into different method calls (`filter` and `map`s) which should make them a bit more understandable.

JiriLojda (changes requested):
Rebase the branch on the #7 and add language ids map into the `ImportContext`.

src/commands/importExportEntities/entities/languages.ts
```diff
@@ -1,11 +1,150 @@
...
+})
+
+const createReplaceIsActiveOperation = (isActive: boolean): LanguageModels.IModifyLanguageData => ({
+  'op': 'replace',
+  'property_name': 'is_active',
+  'value': isActive
+})
+
+
+const updateProjectLanguage = async (client: ManagementClient, projectLanguage: LanguageModels.LanguageModel, importLanguage: LanguageContracts.ILanguageModelContract) => {
```
JiriLojda, line 65: You don't need the `async` keyword here.

src/commands/importExportEntities/entities/languages.ts
```diff
@@ -1,11 +1,150 @@
...
+        return updateProjectLanguage(client, projectLanguage, importLanguage)
+      }
+
+      return client
+        .addLanguage()
+        .withData({
+          name: importLanguage.name,
+          codename: importLanguage.codename,
+          is_active: true,
+          external_id: importLanguage.external_id ?? `${importLanguage.codename}`.slice(0, 50),
```
JiriLojda, line 115: Why do you put codename into the template string, codename is a string on its own. 😄

JiriLojda (approved):
Just a couple of suggestions.

PR #11, opened 2023-12-12
Devrel 1069 import types and snippets

JiriLojda:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/commands/importExportEntities/entities/utils/richText.ts
+++ b/src/commands/importExportEntities/entities/utils/richText.ts
@@ -5,17 +5,17 @@ export const replaceRichTextReferences = (richText: string, context: ImportConte
     .replaceAll(assetRegex, (_, assetId) => {
       const newAssetId = context.assetIdsByOldIds.get(assetId);
       if (!newAssetId) {
-       console.warn(`Found asset id "${assetId}" in rich text that doesn't exist.`);
+        console.warn(`Found asset id "${assetId}" in rich text that doesn't exist.`);
       }
-      
+
       return `${assetAtributeName}="${newAssetId}"`;
     })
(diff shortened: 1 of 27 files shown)


src/commands/importExportEntities/entities/utils/typeElements.ts
```diff
@@ -0,0 +1,260 @@
...
+      return params.builder.textElement({
+        ...typedElement,
+        type: "text",
+        external_id: typedElement.external_id ?? fallbackExternalId,
+        content_group,
+      });
+    }
+    case "url_slug": {
+      const typedElement = element as ContentTypeElements.IUrlSlugElement;
+      if (!("urlSlugElement" in params.builder)) {
```
IvanKiral, line 155: This is kind of hard to follow at first, would not this condition be a good case for typeGuard? also with snippet

src/commands/importExportEntities/entities/utils/richText.ts
```diff
@@ -0,0 +1,27 @@
...
+  richText
+    .replaceAll(assetRegex, (_, assetId) => {
+      const newAssetId = context.assetIdsByOldIds.get(assetId);
+      if (!newAssetId) {
+       console.warn(`Found asset id "${assetId}" in rich text that doesn't exist.`);
+      }
+      
+      return `${assetAtributeName}="${newAssetId}"`;
+    })
+    .replaceAll(itemRegex, (_, itemId) => {
```
IvanKiral, line 13: so Item Id is obtained from the regex group? maybe add comment to clear that up :)

IvanKiral (changes requested):

PR #23, opened 2024-03-07
Add sync skeleton

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


JiriLojda (changes requested):
The high-level structure looks fine to me. I added a few suggestions for possible improvements. You should also update the management SDK to the latest version as it should include better type for the taxonomy patch operation. I will submit an issue to the management SDK repository to improve types for other patch operations as well.

While doing the review, I made a few of your types readonly when they were not. I will commit the changes as they are not very interesting and there is no reason to let you do them again.

src/modules/sync/types/fileContentModel.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  ContentTypeContracts,
+  ContentTypeSnippetContracts,
+  ElementContracts,
+  TaxonomyContracts,
+} from "@kontent-ai/management-sdk";
+
+import { CodenameReference, FixReferences } from "../../../utils/types.js";
+
+export type TaxonomyContract =
```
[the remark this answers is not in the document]
IvanKiral: I like taxonomySyncModel the best

src/modules/sync/generateSyncModel.ts
```diff
@@ -0,0 +1,36 @@
...
+   * The tool might add comment about the datetime and tool's version.
+   */
+
+  // TODO
+  const client = new ManagementClient({
+    environmentId: config.environmentId,
+    apiKey: config.apiKey,
+  });
+
+  const contentTypeModel = generateContentTypeModel(client);
```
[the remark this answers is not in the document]
IvanKiral: sounds good. However I can change the name of functions, but do you also want to change the command name? I think we have model generator, so generate-sync-model as a command sounds good to me :)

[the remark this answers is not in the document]
JiriLojda: Sure, the comment was meant just for the functions.

JiriLojda (changes requested):

src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+) =>
+  types.flatMap(type =>
+    type.elements.filter(element => element.type === "guidelines")
+  ) as unknown as ContentTypeElements.IGuidelinesElement[];
+
+export const getAssetElements = (
+  types: ContentTypeContracts.IContentTypeContract[] | ContentTypeSnippetContracts.IContentTypeSnippetContract[],
+) =>
+  types.flatMap(type =>
+    type.elements.filter(element => element.type === "asset")
```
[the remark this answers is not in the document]
IvanKiral: okay me likey the idea :D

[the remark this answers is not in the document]
IvanKiral: well there is a problem with typegueard of guidelines because ElementContracts.IContentTypeElementContract contains name and ContentTypeElements.IGuidelinesElement does not :/

src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+export const getLinkedItemsElements = (
+  types: ContentTypeContracts.IContentTypeContract[] | ContentTypeSnippetContracts.IContentTypeSnippetContract[],
+) =>
+  types.flatMap(type =>
+    type.elements.filter(element => element.type === "modular_content")
+  ) as unknown as ContentTypeElements.ILinkedItemsElement[];
+
+export const getRequiredAssetsIds = (
+  assetElements: ContentTypeElements.IAssetElement[],
+  portableTextGuidelines: ReadonlyArray<PortableTextObject[]>,
```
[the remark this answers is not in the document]
IvanKiral: well you need to separate item Id from assetId, due to management API call. however I guess i can have function whi returns {assetId: set<string>, itemId: set<string>}  for every content type :D

src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const traverseCallback = (object: PortableTextObject) => {
+    if (object._type === "image") {
+      result.add(object.asset._ref);
+    }
+
+    return object;
+  };
+
+  portableTextGuidelines.map(guideline => guideline.map(g => traversePortableText(g, traverseCallback)));
```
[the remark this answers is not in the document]
IvanKiral: well the documentation of the callback says @param callback
A callback function invoked for each node in the Portable Text structure. It can return a modified version of the node or undefined if no modifications are to be made. I want to traverse the the whole tree and find assetIds and itemIds :D I believe I can traverse it by myself but why would i bother when there is a function that does that for me? I am not sure if you want me to use some reduce with recursion (or one of them by itself).. This PR you are speaking in riddles :D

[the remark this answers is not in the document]
IvanKiral: or you mean transform to json? :)

src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+};
+
+const resolveItemIdsDomHtmlNode: ResolveDomHtmlNodeType = (node, traverse) => {
+  switch (node.tagName) {
+    case "a":
+      return node.attributes["data-item-id"];
+    default: {
+      if (node.children.length) {
```
JiriLojda, line 45: Is it necessary to return `null` when there are no children? I believe it would also work to return an empty array. I you'd return an empty array instead, you would not need this `if`.

src/modules/sync/diff.ts
```diff
@@ -0,0 +1,13 @@
+import { ManagementClient } from "@kontent-ai/management-sdk";
+
+type DiffConfig = DiffOptions;
```
JiriLojda, line 3: I don't understand why this type exists. This is the only usage of the `DiffOptions` type, you can instead rename it to `DiffConfig` and remove this type if you don't like the `DiffOptions` name. 😄

src/modules/sync/generateSyncModel.ts
```diff
@@ -0,0 +1,93 @@
...
+  const allIds = [...contentTypes, ...contentTypeSnippets].reduce<{ assetIds: Set<string>; itemIds: Set<string> }>(
+    (previous, type) => {
+      const ids = getRequiredIds(type);
+
+      return {
+        assetIds: new Set([...previous.assetIds, ...ids.assetIds]),
+        itemIds: new Set([...previous.itemIds, ...ids.itemIds]),
+      };
+    },
+    { assetIds: new Set<string>(), itemIds: new Set<string>() },
```
JiriLojda, line 51: Don't specify the generic type in the `Set` constructor. It should not be necessary.

src/modules/sync/types/fileContentModel.ts
```diff
@@ -0,0 +1,36 @@
...
+  & Readonly<{
+    codename: string;
+    terms: ReadonlyArray<TaxonomySyncModel>;
+  }>;
+
+export type ContentTypeSnippetsSyncModel =
+  & Omit<ContentTypeSnippetContracts.IContentTypeSnippetContract, "id" | "codename" | "last_modified" | "elements">
+  & Readonly<{
+    codename: string;
+    elements: ReadonlyArray<Replace<ElementContracts.IContentTypeElementContract, "codename", string>>;
```
JiriLojda, line 21: What about references inside specific elements (e.g. `allowed_content_types` in a linked items element)?

tests/unit/sync/contentTypeHelpers.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      name: "guidelines",
+      guidelines: guidelinesRichText,
+    },
+    ...linkedItemsElements,
+    ...assetElements,
+  ],
+};
+
+describe("getRequiredAssetsIds", () => {
+  it("get asset ids from elements and guidelines rich text", () => {
```
JiriLojda, line 64: The `describe` and `it` descriptions don't match the tested function.

JiriLojda: I would also prefer to have multiple `it` cases with smaller rich texts for the different cases (item, asset and asset link) instead of one big test case testing all the scenarios.

JiriLojda (changes requested):

src/modules/sync/diff.ts
IvanKiral: at first it combied two types together :D for some reason i did not remove it when deleted the other type :/

src/modules/sync/types/fileContentModel.ts
IvanKiral: well thats a good problem :D `ElementContracts.IContentTypeElementContract` is a type which does contain some "shared" properties (it also contains name which guidelines do not have). however fo example `IAssetElement` does not extend `ElementContracts.IContentTypeElementContract` this (that's why when are doing as unknown as. The types are not so great, so I don't know if that would even help. The solution would be to make type union of all elements but that should be done in SDK in the first place, so for me it does not make sense write it here, when we should change the SDK when the time comes. However, we would still need a function to replace codename? to codename. But i don't see a point doing it now. Or you have some better idea in mind?

src/modules/sync/utils/contentTypeHelpers.ts
JiriLojda: The same applies to the `resolveAssetIdsDomHtmlNode` above. 😉

src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const resolveAssetIdsDomHtmlNode: ResolveDomHtmlNodeType = (node, traverse) => {
+  switch (node.tagName) {
+    case "figure":
+      return node.attributes["data-asset-id"];
+    case "a":
+      return node.attributes["data-asset-id"];
+    default: {
+      if (node.children.length) {
+        return node.children.map(c => traverse(c));
```
JiriLojda, line 33: nit: you can eta-reduce the lambda to `map(traverse)`

src/modules/sync/types/fileContentModel.ts
JiriLojda: Sure, we can create the transformed types of different elements in the tasks that will implement the transformation.

tests/unit/sync/contentTypeHelpers.test.ts
JiriLojda: The name in `describe` still doesn't match the name of the tested function.

tests/unit/sync/contentTypeHelpers.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const ids = getRequiredIds(contentType);
+
+    expect(ids.assetIds).toEqual(new Set(["assetId1"]));
+    expect(ids.itemIds).toEqual(new Set(["itemId1"]));
+  });
+
+  it("obtain ids from asset and linked elements ", () => {
+    const contentType = createContentType([...linkedItemsElements, ...assetElements]);
+    const ids = getRequiredIds(contentType);
```
JiriLojda, line 105: nit: It'd be nice to split the setup part and the act part as you did in the previous cases.

JiriLojda (approved):
I'll approve as the rest of the comments are just minor improvements.

PR #25, opened 2024-03-18
Devrel 1125/transform taxonomies

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


tests/unit/sync/transformTaxonomyGroupModel.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      },
+    ],
+  },
+];
+
+describe("transformTaxonomyGroupsModel", () => {
+  it("transformTaxonomyGroupsModel", () => {
+    const result = transformTaxonomyGroupsModel(input);
+
+    expect(result).toMatchInlineSnapshot(`
```
JiriLojda, line 46: I would prefer creating an adjusted clone of the `input` and check that the `result` equals the `expected` instead of a snapshot test. It would be clearer as to what changes are expected from the `transformTaxonomyGroupsModel`. You can create the `expected` like this:
```ts
const expected = [
  {
    ...input[0],
    id: undefined,
    last_modifed: undefined,
    //...
  },
  //...
];
```

tests/unit/sync/transformTaxonomyGroupModel.test.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, expect, it } from "@jest/globals";
+import { TaxonomyContracts } from "@kontent-ai/management-sdk";
+
+import { transformTaxonomyGroupsModel } from "../../../src/modules/sync/modelTransfomers/taxonomyGroups";
+
+const input: ReadonlyArray<TaxonomyContracts.ITaxonomyContract> = [
+  {
+    id: "taxonomyGroup!Id",
```
JiriLojda, line 8: nit: I suppose this should apply `!` -> `1`, right?

tests/unit/sync/transformTaxonomyGroupModel.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        id: "nestedTermId2",
+        name: "nestedTerm2",
+        codename: "nested_term2",
+        last_modified: "nestedTermModofied2",
+        external_id: "extId_nested_term2",
+        terms: [],
+      },
+    ],
+  },
+] as const;
```
JiriLojda, line 40: The `as const` doesn't do anything here, because you assigned a type to the `input` constant explicitly. If you want to infer the specific type (i.e. if you want this to be a tuple of two groups) and still be a subtype of the `ReadonlyArray<TaxonomyContracts.ITaxonomyContract>`, you must use the `satisfies` keyword like this `as const satisfies ReadonlyArray<TaxonomyContracts.ITaxonomyContract>` and remove the explicit type of the `input` constant.

tests/unit/sync/transformTaxonomyGroupModel.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const expectedOutput = [
```
JiriLojda, line 42: I am not sure about the expected input/output being outside the test case. What if I'll want to add another test case later? Would I be forced to add something like `expectedOutputForTheSecondCase`? I would prefer the data to be part of the `it` callback even though they are pretty long to avoid the problem. The data is not that long after all.

tests/unit/sync/transformTaxonomyGroupModel.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        ...input[1]?.terms[0],
+        id: undefined,
+        last_modified: undefined,
+      },
+    ],
+  },
+];
+
+describe("transformTaxonomyGroupsModel", () => {
+  it("transformTaxonomyGroupsModel", () => {
```
JiriLojda, line 71: Instead of repeating the name of the tested function here, describe what it should do even something as simple as `correctly transforms given groups` is better. The `describe` and `it` combination of functions should help you fluently describe the test cases.

JiriLojda (changes requested):

src/utils/object.ts
```diff
@@ -0,0 +1,4 @@
+type OmitProps<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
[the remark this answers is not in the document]
IvanKiral: To be completely honest, I don't know.. I know about the type and I used it before. For some reason I did that :D

tests/unit/sync/transformTaxonomyGroupModel.test.ts
IvanKiral: I did it because of indentation - wanted to save some space :D for me, there is only one test. When somebody wants to add new test, he might then change it so the test would contain its own expectedOutput :)

JiriLojda: Ok, we can leave it be for now.

JiriLojda (approved):

PR #34, opened 2024-04-22
Devrel 1121/diff command

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


README.md
```diff
@@ -146,6 +146,34 @@ If your export file contains assets bigger than that (they can be uploaded throu
...
+The `generate-sync-model` command is connected to the flow of content model synchronization (sync) between two environments. Its purpose is to generate a folder containing the model of the given Kontent.ai environment for the content types, content type snippets, and taxonomies. The given folder can be used as the source environment for the `sync` command. The generated model follows MAPI format and is stripped of IDs and unnecessary fields (last_modified). References to the other entities (such as snippet in snippet element) are replaced from using `id` to `codename`. Moreover, all entities have filled external ID - if the entity already had an external ID then its value is preserved. Otherwise, the entity gets the value of its ID as an external ID. We also added custom properties `data-asset-codename`, `data-item-codename` and `data-codename` that can be used to reference item or asset inside guidelines' rich-text. These values are removed and transformed to the corresponding ids or external ids during sync. If you want to manually update the values inside this folder, be aware to use the correct values. After the succesfull execution of the command, four files are generated:
```
JiriLojda, line 150: You can split the line into multiple lines in the `.md` file without affecting the visual representation. 😉

JiriLojda: But I would like to split the big chunk of text into smaller parts focused on different parts.
I think we should also highlight that the generated file can be a base for further manual modifications and the model can be completely managed code-first in the file.

README.md
```diff
@@ -146,6 +146,34 @@ If your export file contains assets bigger than that (they can be uploaded throu
...
+- `metadata.json` - contains additional information - not required for sync.
+
+### Usage
+
+```bash
+npx @kontent-ai/data-ops generate-sync-model --environmentId <environment-id> --apiKey <Management-API-key>
+```
+
+## Diff
+The `diff` command compares and prints the diff between two environments. It is possible to enter either two environments parameters (environment ID and MAPI key), or generated folder name (see generate-sync-model), that will be treated as a source content model, with target environment parameters.
```
JiriLojda, line 163: Here I would also highlight that the file can be maintained manually.

README.md
JiriLojda: We should probably also include examples of the generated files. In separate files with links here to avoid bloating readme too much.

JiriLojda (changes requested):

README.md
IvanKiral: I don't think the examples are necessary, just use the command and see for yourself -> it will just add some files that we need to maintain. but I updated the format of the text

JiriLojda (approved):

PR #35, opened 2024-04-22
DEVREL-1140/add webhook entity

pokornyd:
### Motivation

Adds the webhook entity and related operations available via management API.

### Checklist

- [ ] Tests are passing

### How to test


src/commands/importExportEntities/entities/webhooks.ts
```diff
@@ -0,0 +1,44 @@
...
+  importEntities: async (client, entities, context, logOptions) => {
+    await serially(entities.map(webhook => () => {
+      logInfo(logOptions, "verbose", `Importing: webhook ${webhook.id} (${chalk.yellow(webhook.name)})`);
+
+      return client
+        .addWebhook()
+        .withData({
+          name: webhook.name,
+          secret: webhook.secret,
+          url: webhook.url,
```
IvanKiral, line 25: This won't be so easy -> half of the the params are missing  -> you also need to transform references in events from id to codenames

src/commands/importExportEntities/entities/webhooks.ts
```diff
@@ -0,0 +1,44 @@
...
+import chalk from "chalk";
+
+import { logInfo } from "../../../log.js";
+import { serially } from "../../../utils/requests.js";
+import { FixReferences } from "../../../utils/types.js";
+import { EntityDefinition } from "../entityDefinition.js";
+
+type Webhook = FixReferences<WebhookContracts.IWebhookContract>;
+
+export const webhookEntity: EntityDefinition<ReadonlyArray<Webhook>> = {
```
IvanKiral, line 11: you have listed this entity neither in the export command nor the import command -> it won't be called

IvanKiral (changes requested):

src/commands/importExportEntities/entities/webhooks.ts
pokornyd: added transformation logic and missing params

src/commands/importExportEntities/entities/webhooks.ts
pokornyd: added

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,72 @@
...
+  object: Transformable<T>,
+  flatContext: SimplifyContext<any>, // TODO: improve this type
+): T => {
+  /**
+   * this merges all simplified context maps into a single map, which is subsequently used
+   * for lookup of transformed values. if no match is found, original value is returned.
+   *
+   * if the value is an `object`, transforms it recursively.
+   */
+  const flatContextMap = new Map<string, string>(Object.values(flatContext).flatMap(c => [...c]));
```
IvanKiral, line 48: maybe you can use just flat :D even flatmap(c => c) or am missing something?

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const traverseAndReplace = (value: Transformable<T>): Transformable<any> => {
+    switch (typeof value) {
+      case "string":
+        return flatContextMap.get(value) ?? value;
```
IvanKiral, line 53: this is kind of hacky... what if you don't import contentTypes because you excluded them, now you will left old ids in  this triggers

src/commands/importExportEntities/entities/webhooks.ts
```diff
@@ -0,0 +1,60 @@
...
+        return client
+          .addWebhook()
+          .withData({
+            name: webhook.name,
+            secret: webhook.secret,
+            url: webhook.url,
+            enabled: webhook.enabled,
+            delivery_triggers: transformReferences(
+              webhook.delivery_triggers,
+              simplifyContext(context), // TODO: extract only relevant references
```
IvanKiral, line 40: todo

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,72 @@
...
+export const createReference = (params: Params) =>
+  params.newId ? { id: params.newId } : { external_id: `non-existent-${params.entityName}-${params.oldId}` };
+
+/**
+ * extracts top level values from import context. `string:string` pairs are left unchanged,
+ * `string:object` are simplified, with `selfId` string replacing the object as value.
+ *
+ * @param context `ImportContext` object
+ * @param keys specifies which root keys to extract from the context, defaults to all keys
+ * @returns flattened map of string:string pairs
```
IvanKiral, line 20: ok it took me a while but this is not true, I guess?

src/utils/types.ts
```diff
@@ -22,3 +22,33 @@ export type RequiredId<T extends { [key in "id"]?: string }> = Replace<T, "id",
...
+/**
+ * Extracts a transformed `ReadonlyMap` where the values are simplified to the `selfId` property
+ * of the objects if available. Otherwise returns the type as is.
+ *
+ * @template T - The type to transform, e.g. `ReadonlyMap<string, object>`.
+ * @returns A `ReadonlyMap` keyed similarly but with values as the `selfId` of the objects,
+ *          or the original type `T` if `T` is not a `ReadonlyMap` or the objects don't have `selfId`.
+ */
+export type ExtractSelfId<T> = T extends ReadonlyMap<string, infer U>
+  ? U extends { selfId: infer V } ? ReadonlyMap<string, V>
```
IvanKiral, line 35: self id is always string :D I just have feeling this is kind of overengineering

src/utils/types.ts
```diff
@@ -22,3 +22,33 @@ export type RequiredId<T extends { [key in "id"]?: string }> = Replace<T, "id",
...
+  : T
+  : T;
+
+export type Transformable<T> = T extends object ?
```
IvanKiral, line 39: I am not exactly sure why we need this type

src/utils/types.ts
```diff
@@ -22,3 +22,33 @@ export type RequiredId<T extends { [key in "id"]?: string }> = Replace<T, "id",
...
+    & {
+      [P in keyof T]: Transformable<T[P]>;
+    }
+    & { [index: string]: Transformable<any> }
+  : T;
+
+/**
+ * Retypes incorrect typing of filters for webhook actions in SDK (`filters: Array<T>` to `filters: T`)
+ */
+export type FixFiltersType<T> = T extends object ? {
```
IvanKiral, line 49: new sdk version should be released

IvanKiral (changes requested):

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+export const simplifyContext = <T extends keyof ImportContext = keyof ImportContext>(
```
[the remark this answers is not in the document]
IvanKiral: Okay this will probably work as both will point to the zero uuid also, however it would be nice to  check, maybe add comment, that it is dealt with :)

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,72 @@
...
+  context: Partial<ImportContext>,
+  keys?: T[],
+) => {
+  const getSimplifiedMap = (map: ReadonlyMap<string, any>) => new Map([...map].map(([k, v]) => [k, v.selfId ?? v]));
```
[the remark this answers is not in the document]
pokornyd: :/ :/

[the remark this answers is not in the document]
pokornyd: good point, narrowed accordingly.

src/commands/importExportEntities/entities/utils/reference.ts
pokornyd: I think the idea behind c => [...c] was to account for keys that are not arrays/maps, but this is not the case in ImportContext, so I guess it's not required and flat seems like a better solution, good catch.

src/commands/importExportEntities/entities/utils/reference.ts
pokornyd: you mean the return value?

src/utils/types.ts
pokornyd: don't laugh at me :( but yeah, the inference is redundant... I did that on purpose, to see if you review thoroughly... :D

src/utils/types.ts
pokornyd: It's just a recursive definition of a transformable type with index signature, to avoid casting to `any` in the reduce.

src/commands/importExportEntities/entities/webhooks.ts
pokornyd: fixed

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,72 @@
...
+      .filter(([k]) => keysToExtract.includes(k as T))
+      .map(([k, v]) => [k, getSimplifiedMap(v)]),
+  );
+
+  return simpleContext as unknown as SimplifyContext<T>;
+};
+
+export const transformReferences = <T extends object>(
+  object: Transformable<T>,
+  flatContext: SimplifyContext<any>, // TODO: improve this type
```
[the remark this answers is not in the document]
pokornyd: fixed

src/commands/importExportEntities/entities/utils/reference.ts
pokornyd: yup. and the migration will fail, because you don't import contentTypes, even though you have webhooks that rely on them. or am I missing something?

src/utils/types.ts
pokornyd: updated and removed

src/commands/importExportEntities/entities/utils/reference.ts
pokornyd: well... actually... no. I narrowed the type from other comment and now this syntax is explicitly required.

src/utils/types.ts
IvanKiral: Okay, now I see what you wanted to achieve. You wanted to preserve the type when transforming the object. 

This type is used only in `traverseAndReplace`, which is used in `transformReferences` in file `reference.ts`

However, from the type point of view, it's not that easy.  Moreover, from the type point of view, your solution is not kind of correct, therefore you needed to use `any`

`traverseAndReplace` always expects objects and in some cases, you send it to the function which is in runtime string but you retyped it into transformable<string>. Then you are doing type checking whether it is a string, which again in runtime is correct but it is a string, however, the type system is kind of confused (see never type in the below picture)

image

The same is for Array, you said in the function definition that the value would be object, and it is Array (even though I am not sure what happens there as Array is also an object, but I believe it is something workaroundish also)

From the bigger picture point of view, transformReferences takes the object and needs to return the object of same type (it only replaces ids to other ids)

Even though it is nice you tried to make a general solution, in this case I think it would be good enough to keep simpler solution you can make something like this: 

```ts
export const transformReferences = <T extends object, U extends keyof ImportContext>(
  object: T,
  flatContext: SimplifyContext | SimplifyContext<U>,
): T => {
  /**
   * this merges all simplified context maps into a single map, which is subsequently used
   * for lookup of transformed values. if no match is found, original value is returned.
   *
   * if the value is an `object`, transforms it recursively.
   */
  const flatContextMap = new Map<string, string>(Object.values(flatContext).flatMap(c => [...c]));

  const traverseAndReplace = (value: unknown): unknown => {
    if(typeof value === 'string'){
      return flatContextMap.get(value) ?? value
    }

    if(Array.isArray(value)){
      return value.map((item) => traverseAndReplace(item));
    }

    if(value !== null  && typeof value === 'object') {
      return Object.fromEntries(Object.entries(value).map(([key, value]) => [key, traverseAndReplace(value)]));
    }

    return value;
  };

  return traverseAndReplace(object) as T;
};
```
In this state, it is still not very usable everywhere in general, as it might return oldId when it is not found in the context. That could result in more difficulties when debugging - we might see there is an error and then expect that is due to the oldId (that does not exist in the target environment) while the right error would be totally different. However, this is only assumption what might happen, rather what is the actual behavior.

On the other hand, I like that if some references are added (for instance references to collections in webhooks) this solution would handle it, even without us knowing :)

I also wonder if we can use this function in other places than just webhooks (such as content type, options in multiple-choice elements, or even referencing snippet in snippet element), whether it makes sense to have it exported it or have it just inside the webhook entity.

What do you think?

tests/unit/utils/references.test.ts
```diff
@@ -0,0 +1,72 @@
+import { describe, expect, it } from "@jest/globals";
+
+import {
+  simplifyContext,
+  transformReferences,
+} from "../../../src/commands/importExportEntities/entities/utils/reference";
+import { ImportContext } from "../../../src/commands/importExportEntities/entityDefinition";
+
+const emptyMap = new Map();
+const contextTemplate: Partial<ImportContext> = {
```
IvanKiral, line 10: Nitpick: I would use Pick<ImportContext, "collectionIdsByOldIds" | "contentTypeSnippetContextByOldIds">

IvanKiral (changes requested):

src/commands/importExportEntities/entityDefinition.ts
```diff
@@ -95,4 +96,16 @@ export type ImportContext = Readonly<{
...
 
+/**
+ * Allows extracting a subset of `ImportContext` based on keys specified. Nested values in maps are simplified
+ * (replaces object by its `selfId`).
+ *
+ * @template K - A subset of keys from `ImportContext` (defaults to all keys).
+ *
+ * @example type SimplifiedContext = ExtractRootEntityMaps<'workflowIdsByOldIds' | 'collectionIdsByOldIds'>;
+ */
+export type SimplifyContext<K extends keyof ImportContext = keyof ImportContext> = {
```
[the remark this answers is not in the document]
pokornyd: fixed

tests/integration/importExport/utils/compare.ts
```diff
@@ -643,6 +645,77 @@ const prepareRoleReferences: PrepareReferencesFnc<RoleContracts.IRoleContract> =
...
+    last_modified: "-",
+    health_status: "-",
+    delivery_triggers: {
+      ...webhook.delivery_triggers,
+      content_item: webhook.delivery_triggers.content_item
+        ? {
+          ...webhook.delivery_triggers.content_item,
+          filters: webhook.delivery_triggers.content_item.filters
+            ? {
+              ...webhook.delivery_triggers.content_item.filters,
```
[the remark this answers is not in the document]
pokornyd: found only one such occurrence, removed it. if I missed any, let me know :)

tests/unit/utils/references.test.ts
pokornyd: fixed

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,73 @@
...
+      .filter(([k]) => keysToExtract.includes(k as T))
+      .map(([k, v]) => [k, getSimplifiedMap(v)]),
+  );
+
+  return simpleContext as unknown as SimplifyContext<T>;
+};
+
+export const transformReferences = <T extends object, U extends keyof ImportContext>(
+  object: Transformable<T>,
+  flatContext: SimplifyContext | SimplifyContext<U>,
```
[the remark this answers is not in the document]
pokornyd: split the methods into two, combined them in a higher order

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,78 @@
...
+}>;
+
+export const createReference = (params: Params) =>
+  params.newId ? { id: params.newId } : { external_id: `non-existent-${params.entityName}-${params.oldId}` };
+
+/**
+ * extracts top level values from import context. `string:string` pairs are left unchanged,
+ * `string:object` are simplified, with `selfId` string replacing the object as value.
+ */
+const simplifyContext = <K extends keyof ImportContext = keyof ImportContext>(
```
IvanKiral, line 16: simplify with flattern

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,78 @@
...
+  keys?: K[],
+) => flattenContext(simplifyContext(context, keys));
+
+export const transformReferences = <T extends object>(
+  object: T,
+  flatContext: Map<string, string>,
+): T => {
+  /**
+   * this merges all simplified context maps into a single map, which is subsequently used
+   * for lookup of transformed values. if no match is found, original value is returned.
```
IvanKiral, line 55: probably this comment is not necessary

src/utils/types.ts
```diff
@@ -40,3 +40,23 @@ export type SuperiorPick<T, K extends keyof T> = T extends any ? {
...
+
+/**
+ * Extracts a transformed `ReadonlyMap` where the values are simplified to the `selfId` property
+ * of the objects if available. Otherwise returns the type as is.
+ *
+ * @template T - The type to transform, e.g. `ReadonlyMap<string, object>`.
+ * @returns A `ReadonlyMap` keyed similarly but with values as the `selfId` of the objects,
+ *          or the original type `T` if `T` is not a `ReadonlyMap` or the objects don't have `selfId`.
+ */
+export type ExtractSelfId<T> = T extends ReadonlyMap<string, infer U>
```
IvanKiral, line 52: this type is no longer used anywhere, let's delete it :)

src/utils/types.ts
```diff
@@ -40,3 +40,23 @@ export type SuperiorPick<T, K extends keyof T> = T extends any ? {
...
+  ? U extends { selfId: string } ? ReadonlyMap<string, string>
+  : T
+  : T;
+
+export type Transformable<T> = T extends object ?
+    & {
+      [P in keyof T]: Transformable<T[P]>;
+    }
+    & { [index: string]: Transformable<any> }
```
IvanKiral, line 61: `Transformable<unknown>`

src/commands/importExportEntities/entities/utils/reference.ts
```diff
@@ -0,0 +1,77 @@
...
+  context: Partial<ImportContext>,
+  keys?: K[],
+) => flattenContext(simplifyContext(context, keys));
+
+export const transformReferences = <T extends object>(
+  object: T,
+  flatContext: Map<string, string>,
+): T => {
+  /**
+   * this merges all simplified context maps into a single map, which is subsequently used
```
IvanKiral, line 54: this function does not merge... it gets merged context maps as a parameter.

src/commands/importExportEntities/entityDefinition.ts
```diff
@@ -95,4 +95,16 @@ export type ImportContext = Readonly<{
...
 
+/**
+ * Allows extracting a subset of `ImportContext` based on keys specified.
+ * string:object maps are simplified to string:string.
+ *
+ * @template K - A subset of keys from `ImportContext` (defaults to all keys).
+ *
+ * @example type SimplifiedContext = SimplifyContext<'workflowIdsByOldIds' | 'collectionIdsByOldIds'>;
+ */
+export type SimplifyContext<K extends keyof ImportContext = keyof ImportContext> = {
```
IvanKiral, line 106: nitpick: I would rename this into SimplifiedContent to not use verb.. it kind of suggest me to be a function instead of a type

IvanKiral (changes requested):
close to  finishing

src/utils/types.ts
pokornyd: probably got introduced with the rebase, removed.

src/utils/types.ts
pokornyd: this type shouldn't be here at all, it's no longer used.

src/commands/importExportEntities/entities/utils/reference.ts
pokornyd: also outdated from the rebase. fixed

src/commands/importExportEntities/entityDefinition.ts
pokornyd: again, unused type, removed :)

IvanKiral (approved):

PR #38, opened 2024-04-25
Fix external id in import

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/commands/import.ts
```diff
@@ -124,13 +124,18 @@ const importEntities = async (params: ImportEntitiesParams) => {
     logInfo(params, "standard", `Importing: ${chalk.yellow(def.name)}`);
 
     try {
-      context = await root.entryData(`${def.name}.json`)
+      context = await root.entryData(`${def.filename}.json`)
```
JiriLojda, line 127: You also need to change this in the export command.

JiriLojda (changes requested):

src/commands/export.ts
```diff
@@ -131,7 +131,9 @@ const exportEntities = async (params: ExportEntitiesParams): Promise<void> => {
     } catch (err) {
       logError(
         params,
-        `Failed to export an entity ${chalk.red(def.name)} due to error ${JSON.stringify(err)}. Stopping export...`,
+        `Failed to export an entity ${chalk.red(def.name)} due to error ${
```
JiriLojda, line 134: Put the `message` property here as well to be consistent with the import command.

JiriLojda (changes requested):

JiriLojda (changes requested):
You should also use the readable name when displaying the "Exporting: ..." and "Cleaning: ..." messages.

JiriLojda (approved):

PR #42, opened 2024-05-07
Devrel 1130/sync command

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/utils/error.ts
+++ b/src/utils/error.ts
@@ -1,3 +1,40 @@
+import { SharedModels } from "@kontent-ai/management-sdk";
+import { isAxiosError } from "axios";
+
 export const throwError = (msg: string) => {
   throw new Error(msg);
 };
+
+export const skipKontentErrors = (errorCodesToSkip: ReadonlyArray<number>) =>
+  handleKontentErrors(e => {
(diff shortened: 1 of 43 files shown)


src/utils/error.ts
```diff
@@ -1,3 +1,47 @@
...
   throw new Error(msg);
 };
+
+export type KontentError = Readonly<{
+  kontentErrorCode: number;
+  message: string;
+  validationMessages: ReadonlyArray<string>;
+}>;
+
+export const skipKontentErrors = (errorCodesToSkip: ReadonlyArray<number>) =>
```
IvanKiral, line 14: same with this funciton, it is not used anywhere, but might be possibly useful in future :D was this the intention?

JiriLojda (changes requested):

src/utils/error.ts
JiriLojda: Yeah, I think it may be useful in the future. For now, I think we can use it to ignore the 404 error when we are checking for existence of items and assets in validation.

JiriLojda (approved):
We will fix the integration tests after rebasing on main

PR #57, opened 2024-07-04
Devrel 1030/migrations

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/migrations/utils/errUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { logError, LogOptions } from "../../../log.js";
+
+export type Err = { err: unknown };
+
+export type ErrLike<T> = T | Err;
+
+export const isErr = (entity: ErrLike<unknown>): entity is Err =>
+  typeof entity === "object" && entity !== null && "err" in entity;
```
JiriLojda, line 8: You can use `zod` here as well, right?

src/utils/types.ts
```diff
@@ -40,3 +40,7 @@ export type SuperiorPick<T, K extends keyof T> = T extends any ? {
  * Original Omit type extended to work on Union type objects
  */
 export type SuperiorOmit<T, K extends keyof any> = T extends any ? SuperiorPick<T, Exclude<keyof T, K>> : never;
+
+export type MakeUnion<Object extends object> = keyof Object extends infer Keys
```
JiriLojda, line 44: I am not quite sure I understand this type. I seems to make every property of the provided object required and remove `undefined` from its values. But that doesn't fit the name of the type so I am confused.

tests/unit/migrations/migrationUtils.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  it("Correctly filters migrations for all parameter", () => {
+    const filteredMigrations = filterMigrations(migrations, { all: true });
+
+    expect(filteredMigrations).toStrictEqual(migrations);
+  });
+
+  it("Correctly filters migrations for range parameter", () => {
+    const filteredMigrations = filterMigrations(migrations, { range: { from: 2, to: new Date("2024-03-07") } });
+
+    expect(filteredMigrations).toStrictEqual([migrations[1], migrations[2], migrations[3]]);
```
JiriLojda, line 46: I know this is works as intended, but I find it quite unintuitive that the "Migration3" gets picked up as well. Maybe we should consider refusing ranges combined of a number and a date. What do you think?

tests/unit/migrations/migrationUtils.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      { logLevel: "none" },
+    );
+
+    const migrationStatusWithoutDate = migrationStatus.status.map(s => ({ ...s, time: undefined }));
+    const expectedResult = [
+      createMigrationStatus(migrations[0].name, migrations[0].module.order, true, "run"),
+      createMigrationStatus(migrations[1].name, migrations[1].module.order, true, "run"),
+    ].map(s => ({ ...s, time: undefined }));
+
+    expect(migrationStatusWithoutDate).toStrictEqual(expectedResult);
```
JiriLojda, line 116: I would be nice to check that the `run` functions actually got called with a properly setup client. This is, after all, the most important part of the whole migration command. 😄

tests/unit/migrations/range.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe("parseRange", () => {
+  it("should return an error for bad range format", () => {
+    expect(parseRange("T2023-01-01")).toEqual({ err: "Bad range format" });
+    expect(parseRange("T202:")).toEqual({ err: "T202 is not a valid order format." });
+    expect(parseRange("Tfdsa:")).toEqual({ err: "Tfdsa is not a valid order format." });
+    expect(parseRange("T2024:T202")).toEqual({ err: "T202 is not a valid order format." });
+    expect(parseRange("T2024-0:T202")).toEqual({ err: "T2024-0 is not a valid order format." });
+    expect(parseRange("T2024-04:2024-05")).toEqual({ err: "2024-05 is not a valid order format." });
+    expect(parseRange("T2023-01-01:2023-01-01:2023-01-01")).toEqual({ err: "Bad range format" });
```
JiriLojda, line 13: A more common format for parametrized tests is like this:
```ts
[
  { range: "T2023-01-01", err: "Bad range format" },
  //...
]
  .forEach(({ range, err}) => {
    it('should return "err" for range "range"', () => {
      expect(parseRange(range)).toStrictEqual({ err });
    });
});
```

tests/unit/migrations/range.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  it("should return an error for invalid date formats", () => {
+    expect(parseRange("T2023-13-01:T2023-01-02")).toEqual({ err: "T2023-13-01 is invalid Date." });
+    expect(parseRange("T2023-01-01:T2023-01-32")).toEqual({ err: "T2023-01-32 is invalid Date." });
+  });
+});
+
+describe("validateRange", () => {
+  it("should return an error for invalid order format", () => {
+    const err = { err: "Left sight of range can't be bigger than right side!" };
```
JiriLojda, line 40: "Left sight"? 😄

tests/unit/migrations/range.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it("should return an error for invalid date formats", () => {
+    expect(parseRange("T2023-13-01:T2023-01-02")).toEqual({ err: "T2023-13-01 is invalid Date." });
+    expect(parseRange("T2023-01-01:T2023-01-32")).toEqual({ err: "T2023-01-32 is invalid Date." });
+  });
+});
+
+describe("validateRange", () => {
+  it("should return an error for invalid order format", () => {
```
JiriLojda, line 39: If it only validates that the left side is lower than the right side, it would be nice to mention it in the test name.

tests/unit/migrations/range.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(parseRange("T2023-01-01")).toEqual({ err: "Bad range format" });
+    expect(parseRange("T202:")).toEqual({ err: "T202 is not a valid order format." });
+    expect(parseRange("Tfdsa:")).toEqual({ err: "Tfdsa is not a valid order format." });
+    expect(parseRange("T2024:T202")).toEqual({ err: "T202 is not a valid order format." });
+    expect(parseRange("T2024-0:T202")).toEqual({ err: "T2024-0 is not a valid order format." });
+    expect(parseRange("T2024-04:2024-05")).toEqual({ err: "2024-05 is not a valid order format." });
+    expect(parseRange("T2023-01-01:2023-01-01:2023-01-01")).toEqual({ err: "Bad range format" });
+  });
+
+  it("should parse ranges correctly", () => {
```
JiriLojda, line 16: I don't see any range using hours, minutes or seconds. We should test that as well.

tests/unit/migrations/statusUtils.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    const result = updateEnvironmentStatus(originalEnvStatus, updateEnvStatus).map(s => ({ ...s, time: undefined }));
+
+    const expectedResult = [originalEnvStatus[0], updateEnvStatus[0], originalEnvStatus[2], updateEnvStatus[1]]
+      .map(s => ({ ...s, time: undefined }));
+
+    expect(result).toStrictEqual(expectedResult);
+  });
+});
+
+describe("defaultReadStatus", () => {
```
JiriLojda, line 40: What about "defaultWriteStatus"?

src/commands/migrations/add/add.ts
```diff
@@ -0,0 +1,98 @@
...
+    builder: yargs =>
+      yargs
+        .option("name", {
+          alias: "n",
+          describe: "Name of the migration.",
+          demandOption: "You need to provide name of the migration.",
+          type: "string",
+        })
+        .option("migrationsFolder", {
+          alias: "p",
```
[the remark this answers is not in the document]
IvanKiral: Hmm at first it was path, then i forgot to change the alias.. still not convinced about the name of the parameter. what do you think?

src/modules/migrations/utils/errUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { logError, LogOptions } from "../../../log.js";
+
+export type Err = { err: unknown };
+
+export type ErrLike<T> = T | Err;
```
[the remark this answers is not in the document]
IvanKiral: I like WithErr

src/utils/types.ts
IvanKiral: it makes union type from a type of one object. Each item in union has one of the property so from {a: string, b: string} it would create {a: string} | {b:string}, I made that type as a return type to the function that checks whether the user gave the right parameters (name, all, next or range)

tests/unit/migrations/migrationUtils.test.ts
IvanKiral: Why? I think it is nice if you had a folder of combined orders. I believe that the ordering with dates sucks, but I don;t really wanna delete, because somebody might use it :D

tests/unit/migrations/statusUtils.test.ts
IvanKiral: the defaultWriteStatus is a very simple function which uses just node functions that i don't think needs to be tested.. Howver the function was originally name createDefaultSaveStatus, so I renamed it do createDefaultWriteStatus :)

tests/unit/migrations/range.test.ts
IvanKiral: applied to all tests in range.test.ts

src/modules/migrations/utils/migrationUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export const executeMigrations = async (
+  migrations: ReadonlyArray<Migration>,
+  client: ManagementClient,
+  options: ExecuteMigrationOptions,
+  logOptions: LogOptions,
+) => {
+  const status: Array<MigrationStatus> = [];
+
+  for (const migration of migrations) {
```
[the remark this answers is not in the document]
IvanKiral: fixed by adding function seriallyReduce

.eslintrc.json
```diff
@@ -13,6 +13,14 @@
           {
             "selector": "MemberExpression[object.name='console'][property.name=/^(log|warn|error|info|trace)$/]",
             "message": "Don't log into the console directly. Use one of the functions from the log.ts file (logInfo, logWarning, logError) to ensure provided --logLevel is respected."
+          }, 
+          {
+            "selector": "CallExpression[callee.name='createManagementClient']",
+            "message": "Use provided create client in src/utils/client.ts"
```
JiriLojda, line 19: What about a more specific `Use the createClient function from src/utils/client.ts instead.`?

src/commands/migrations/run/run.ts
```diff
@@ -0,0 +1,238 @@
...
+    `⚠ Running this operation may result in irreversible changes to your environment ${params.environmentId}.\nOK to proceed y/n? (suppress this message with --skipConfirmation parameter)\n`,
+  );
+
+  const confirmed = !params.skipConfirmation ? await requestConfirmation(warningMessage) : true;
+
+  if (!confirmed) {
+    process.exit(0);
+  }
+
+  const client = createManagementClient({
```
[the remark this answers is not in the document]
JiriLojda: Thank you for adding the rule. 👍

src/commands/migrations/add/add.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    logInfo(params, "standard", `Folder ${chalk.blue(resultFolderPath)} has been created successfully.`);
+  }
+
+  const currentDate = new Date();
+
+  const migrationName = getMigrationName(
+    params.name,
+    params.type as "js" | "ts",
+    params.timestamp ? formatDateForFileName(currentDate) : undefined,
```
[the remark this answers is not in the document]
JiriLojda: np, just an idea 👍

src/utils/types.ts
JiriLojda: Oh, I understand now. The new naming helps, but a small docs would help even more. Maybe with one example. 😉

JiriLojda (approved):
Now there are just some minor things that can be improved.

src/utils/requests.ts
```diff
@@ -9,6 +9,22 @@ export const serially = async <Fetchers extends ReadonlyArray<() => Promise<unkn
...
 
+export const seriallyReduce = async <ReturnType>(
+  fetchers: ReadonlyArray<(prevValue: ReturnType) => Promise<ReturnType>>,
+  initialValue: ReturnType,
+): Promise<ReturnType> => {
+  if (!fetchers.length) {
+    return initialValue;
+  }
+
+  let result = await fetchers[0]?.(initialValue) as ReturnType;
```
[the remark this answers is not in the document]
IvanKiral: muchas gracias, much better :D

PR #61, opened 2024-08-08
Add web spotlight entity

JiriLojda:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/sync/utils/getContentModel.ts
```diff
@@ -17,19 +27,20 @@ export const readContentModelFromFolder = async (folderName: string): Promise<Fi
...
-  >;
+  ) as ReadonlyArray<ContentTypeSnippetsSyncModel>;
   const taxonomyGroups = JSON.parse(
     await fs.readFile(`${folderName}/${taxonomiesFileName}`, "utf8"),
-  ) as unknown as ReadonlyArray<
-    TaxonomySyncModel
-  >;
+  ) as ReadonlyArray<TaxonomySyncModel>;
+
+  const webSpotlight = JSON.parse(
```
IvanKiral, line 35: maybe we can use zod for this? 👀 If you don't want to do that in this PR, maybe we can create a task for that :)

tests/unit/syncModel/webSpotlightTransformer.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  it("correctly transforms enabled web spotlight with existing root type", () => {
+    const environmentModel = createEnvironmentModel(true, "typeId1");
+    const expectedOutput = { enabled: true, root_type: { codename: "type_codename_1" } };
+
+    const result = transformWebSpotlightModel(environmentModel);
+
+    expect(result).toEqual(expectedOutput);
+  });
+
+  it("correctly transforms enabled web spotlight with non-existing root type", () => {
```
IvanKiral, line 47: I would rather test when the web spotlight is deactivated but root_type isn't.. This one can happen, when you deactivate the web_spotlight :)

Moreover I am not sure, whether you can have enabled web_spotlight without any type

tests/unit/utils/requests.test.ts
```diff
@@ -26,7 +26,7 @@ describe("request utils", () => {
     });
   });
 
-  describe.only("seriallyReduce", () => {
+  describe("seriallyReduce", () => {
```
IvanKiral, line 29: thanks bro

IvanKiral (changes requested):

src/modules/sync/utils/getContentModel.ts
JiriLojda: I'll create a task to use zod to validate all the input files (both for sync and import)

JiriLojda: https://kontent-ai.atlassian.net/browse/DEVREL-1202

src/modules/sync/sync.ts
```diff
@@ -43,14 +43,17 @@ export const sync = async (client: ManagementClient, diff: DiffModel, logOptions
   logInfo(logOptions, "standard", "Updating content types and adding their references");
   await updateContentTypesAndAddReferences(client, diff.contentTypes);
 
+  logInfo(logOptions, "standard", "Updating web spotlight");
```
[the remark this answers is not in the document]
JiriLojda: I am not sure what you mean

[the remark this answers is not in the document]
IvanKiral: I mean explaining the order of these operations. As we first create all the types, when enabling Web Spotlight we don't have to worry that the type we are referencing won't exist. Due to this we always use the option when we provide the `root_type` parameter. We never use `null` when enabling Web Spotlight. Ideally, we would also update the diagram in the images :D

IvanKiral (changes requested):

IvanKiral (approved):

PR #62, opened 2024-08-12
expose apis

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


tests/integration/sync/syncModel.test.ts
```diff
@@ -53,14 +54,17 @@ describe("Sync model of two environments with credentials", () => {
...
-      await expectSameSyncEnvironments(environmentId, SYNC_TARGET_TEST_ENVIRONMENT_ID);
+      await syncModelRun({
+        sourceEnvironmentId: SYNC_TARGET_TEST_ENVIRONMENT_ID,
+        sourceApiKey: API_KEY,
+        targetEnvironmentId: environmentId,
+        targetApiKey: API_KEY,
+        verbose: true,
+      });
+
+      await await expectSameSyncEnvironments(environmentId, SYNC_TARGET_TEST_ENVIRONMENT_ID);
```
JiriLojda, line 67: One `await` is enough. 😄

JiriLojda (changes requested):
I would prefer to list all the exported symbols in the `public.ts` instead of the tree of barrel files. It would be easier to see the public interface of the package and harder to accidentally expose something unwanted.

src/modules/sync/diffEnvironments.ts
```diff
@@ -0,0 +1,82 @@
...
+    targetEnvironmentId: string;
+    targetApiKey: string;
+  }>
+  & (
+    | Readonly<{ folderName: string }>
+    | Readonly<{ sourceEnvironmentId: string; sourceApiKey: string }>
+  )
+  & (
+    Readonly<{
+      advanced: boolean;
```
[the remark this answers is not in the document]
JiriLojda: You can wrap the whole type in the `Readonly` type to reduce the amount of the "readonly" keywords. 😉

src/modules/importExport/utils/includeExclude.ts
```diff
@@ -0,0 +1,12 @@
+import { match, P } from "ts-pattern";
+
+export type IncludeExclude<Choices extends string> =
+  | { include: ReadonlyArray<Choices>; exclude?: undefined }
+  | { exclude: ReadonlyArray<Choices>; include?: undefined } // Exclude is optional because it's possible to use the type without
+  | {};
```
JiriLojda, line 6: Don't use this type (`{}`). Anything is assignable to it (e.g. try this `const shit: IncludeExclude<"a" | "b"> = 88;`, it doesn't show an error). You shouldn't need it here. Just make `include`/`exclude` optional in both cases. It will even work with the `ts-pattern` condition bellow.

JiriLojda: And make the type readonly, please. 😄

src/commands/migrations/run/run.ts
```diff
@@ -126,110 +110,80 @@ export const register: RegisterCommand = yargs =>
...
+  });
+};
 
-  if (migrationsStatus.error) {
-    process.exit(1);
-  }
+const resolveParams = (
+  params: RunMigrationsCliParams,
+): WithErr<RunMigrationsParams> => {
+  const baseObj = { next: undefined, name: undefined, range: undefined, all: undefined };
```
JiriLojda, line 174: `emptyParams`?

src/commands/importExport/import.ts
```diff
@@ -85,88 +42,47 @@ export const register: RegisterCommand = yargs =>
...
-  oldAssetCodenamesByIds: new Map(),
-  contentTypeSnippetContextByOldIds: new Map(),
-  contentItemContextByOldIds: new Map(),
-  oldContentItemCodenamesByIds: new Map(),
-  contentTypeContextByOldIds: new Map(),
-  workflowIdsByOldIds: new Map(),
-  workflowStepsIdsWithTransitionsByOldIds: new Map(),
-  spaceIdsByOldIds: new Map(),
-});
+const resolveParams = (params: ImportEnvironmentCliParams): ImportEnvironmentParams =>
```
JiriLojda, line 84: What about putting this into the `includeExclude.ts` as something like `resolveIncludeExcludeFromCli`?

JiriLojda (changes requested):

src/modules/importExport/utils/includeExclude.ts
JiriLojda: You forgot a comment there that doesn't seem relevant anymore.

JiriLojda (approved):

PR #67, opened 2024-09-02
Sync collections

IvanKiral:
> [!WARNING] 
> waiting for asset folders to be merged

### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/sync/printDiff.ts
```diff
@@ -23,6 +23,12 @@ export const printDiff = (diffModel: DiffModel, logOptions: LogOptions) => {
     logInfo(logOptions, "standard", "No asset folders to update.");
   }
 
+  if (diffModel.collections.length) {
+    diffModel.assetFolders.forEach(op => printPatchOperation(op, logOptions));
```
JiriLojda, line 27: Probably a copy/paste mistake. 😄

src/modules/sync/sync/collections.ts
```diff
@@ -0,0 +1,33 @@
...
+    .toPromise();
+};
+
+const transformCollectionsReferences = (operation: PatchOperation): CollectionModels.ISetCollectionData => {
+  const pathParts = operation.path.split("/");
+  const propertyName = pathParts[pathParts.length - 1];
+  const codename = getTargetCodename(operation);
+
+  return {
+    ...operation,
```
JiriLojda, line 23: Use the `omit` function instead of setting the unwanted properties to `undefined`.

src/modules/sync/utils/fetchers.ts
```diff
@@ -97,3 +97,9 @@ export const fetchAssetFolders = (client: ManagementClient) =>
     .listAssetFolders()
     .toPromise()
     .then(res => res.rawData.folders);
+
+export const fetchCollections = (client: ManagementClient) =>
+  client
+    .listCollections()
+    .toPromise()
+    .then(res => res.data.collections.map(c => c._raw));
```
JiriLojda, line 105: You can just get `res.rawData.collections` and avoid the `map`. It's not super important, but the raw array is already there, so why not use it?

tests/integration/sync/syncModel.test.ts
```diff
@@ -99,10 +102,23 @@ describe.concurrent("Sync environment from folder", () => {
       .then(stats => stats.isFile())
       .catch(() => false);
 
+    const webSpotlightJsonExists = await fsPromises.stat(`${folderPath}/${webSpotlightFileName}`)
```
JiriLojda, line 105: Thank you, I didn't notice this when adding the entities. Can we make the check automatically for all the constants in the file?

JiriLojda (changes requested):

tests/integration/sync/syncModel.test.ts
```diff
@@ -89,20 +85,15 @@ describe.concurrent("Sync environment from folder", () => {
...
-    expect(snippetJsonExists).toEqual(true);
-    expect(taxonomiesJsonExists).toEqual(true);
+
+    const result = await Promise.all(
+      Object.entries(fileNames).map(([, filename]) =>
+        fsPromises.stat(`${folderPath}/${filename}`).then(stats => stats.isFile()).catch(() => false)
+      ),
+    );
+
+    expect(result.every(res => res)).toEqual(true);
```
JiriLojda, line 96: Thank you for the improvement. This will always check that all the files are present. But I suggest we also improve the error message in case it fails. We can create an object with keys set to fileNames and values would be `true`/`false` depending on the file's existence. Then, we can do `expect(filesExistence).toStrictEqual(allFilesExist)`. Then the error message would tell you precisely which files are missing.

src/modules/sync/validation.ts
```diff
@@ -39,6 +39,17 @@ export const validateDiffedModel = async (
     `Content type (codename: ${typeCodename}) is being used and can't be removed (In given environment exists content items of given content type)`
   );
 
+  const collectionsUsage = await getUsedCollectionsCodenames(
+    client,
+    new Set(
+      diffedModel.collections.filter(c => c.op === "remove").map(c => getTargetCodename(c)).filter(notNullOrUndefined),
```
JiriLojda, line 45: nit: You can eta-reduce the `map` argument. (`map(getTargetCodename)`)

JiriLojda (changes requested):

PR #70, opened 2024-09-12
Devrel 1205/partial sync

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/sync/syncModelRun.ts
```diff
@@ -75,11 +130,14 @@ export const getDiffModel = async (
     params,
   );
 
+  const filteredSourceModel = filterModel(sourceModel, params.entities);
+  const filteredTargetModel = filterModel(transformedTargetModel, params.entities);
```
JiriLojda, line 134: It'd be better not to even fetch the entities that are not supposed to be synced. To avoid unnecessary fetch of all the entities when one wants to sync just collections for example.

src/modules/sync/syncModelRun.ts
```diff
@@ -1,22 +1,64 @@
...
+export type SyncEntities = Partial<
+  Expect<ExpectedSyncEntities, {
+    contentTypes: (type: ContentTypeSyncModel) => boolean;
+    contentTypeSnippets: (snippet: ContentTypeSnippetsSyncModel) => boolean;
+    taxonomies: (taxonomy: TaxonomySyncModel) => boolean;
+    assetFolders: (assetFolder: AssetFolderSyncModel) => boolean;
+    collections: (collection: CollectionSyncModel) => boolean;
+    spaces: (space: SpaceSyncModel) => boolean;
+    languages: (language: LanguageSyncModel) => boolean;
+    workflows: (workflow: WorkflowSyncModel) => boolean;
```
JiriLojda, line 52: Add a jsdoc warning that syncing workflows removes their role restrictions.

src/commands/syncModel/run/README.md
```diff
@@ -66,6 +66,35 @@ const params: SyncModelRunParams = {
 await syncModelRun(params);
 ```
 
+### Partial Sync
+
+The tool allows you to control which data from your environment gets synced. Using the CLI command, you can filter data by entity type with the `--include` or `--exclude` parameters. Additionally, you can specify an array of codenames to be synced using the `--codenames` parameter, which will apply the codename filter across all synced entities. For more advanced filtering, you can use the `filter` parameter through [programmatic sync](#sync-model-programmatically), where you can define custom filter predicate functions for each entity. These functions are optional, and if not provided, the entity will be synced by default.
```
JiriLojda, line 71: This is now different. Only the `entities` and `codenames` parameters are present, right?

src/commands/syncModel/run/README.md
```diff
@@ -66,6 +66,35 @@ const params: SyncModelRunParams = {
 await syncModelRun(params);
 ```
 
+### Partial Sync
```
JiriLojda, line 69: We should probably incorporate it into the readme instead of having the `Partial sync` section since the partial sync is not sort of a default.

src/modules/sync/diff.ts
```diff
@@ -156,7 +156,10 @@ const createDiffModel = <Entity extends Readonly<{ codename: string }>>(
 const getLanguageDiffModel = (
   sourceLanguages: ReadonlyArray<LanguageSyncModel>,
   targetLanguages: ReadonlyArray<LanguageSyncModel>,
-) => {
+): DiffModel["languages"] => {
+  if (sourceLanguages.length === 0 && targetLanguages.length === 0) {
```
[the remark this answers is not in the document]
IvanKiral: yes, but this condition says when both languages are empty.. that's the only case when the we are not syncing languages

src/modules/sync/generateSyncModel.ts
```diff
@@ -219,3 +220,19 @@ type FileContentWithMetadata =
...
+export const filterModel = (model: FileContentModel, entities: SyncEntities): FileContentModel => ({
+  contentTypes: filterEntities(model.contentTypes, entities.contentTypes),
+  contentTypeSnippets: filterEntities(model.contentTypeSnippets, entities.contentTypeSnippets),
+  taxonomyGroups: filterEntities(model.taxonomyGroups, entities.taxonomies),
+  assetFolders: filterEntities(model.assetFolders, entities.assetFolders),
+  collections: filterEntities(model.collections, entities.collections),
+  spaces: filterEntities(model.spaces, entities.spaces),
+  languages: filterEntities(model.languages, entities.languages),
+  // when turned syncning web-spotlight off, give default object - diff will find no patch operations for same objects.
+  webSpotlight: entities.webSpotlight ? model.webSpotlight : { enabled: false, root_type: { codename: "no-sync" } },
```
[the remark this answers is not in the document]
IvanKiral: it doesn't set to anything. both target and source would have the same object and the diff will return empty operations

src/modules/sync/syncModelRun.ts
```diff
@@ -1,22 +1,64 @@
...
+  WorkflowSyncModel,
+} from "./types/syncModel.js";
 import {
   getSourceItemAndAssetCodenames,
   getTargetContentModel,
   readContentModelFromFolder,
 } from "./utils/getContentModel.js";
 import { validateDiffedModel, validateSyncModelFolder } from "./validation.js";
 
+export const SyncEntityChoices = [
```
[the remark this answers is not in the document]
IvanKiral: just a typo, updated (y)

src/modules/sync/syncModelRun.ts
IvanKiral: how would this work? we usually fetch all units from an entity, as we do not have filtering option in Management SDK. What I can do is that those undefined properties in entities option I can prevent those from fetching.. However still need to filter here. What is more some might be still needed for transformation (ids -> to codenames), for example types needs snippets and so on

src/commands/syncModel/run/run.ts
```diff
@@ -53,6 +57,19 @@ export const register: RegisterCommand = yargs =>
...
         })
+        .option("entities", {
+          alias: "e",
+          type: "array",
+          choices: SyncEntityChoices,
+          describe: `Sync specified entties. Allowed entities are: ${SyncEntityChoices.join(", ")}`,
+          demandOption: "You need to provide the what entities to sync.",
+          conflicts: "exclude",
+        })
+        .option("codenames", {
```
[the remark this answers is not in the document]
IvanKiral: is that possible? because I am not sure :D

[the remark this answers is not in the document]
JiriLojda: It is not, but we can have an option for each of the entities accepting an array of codenames with empty array meaning all the entities. What do you think?

src/modules/sync/diff.ts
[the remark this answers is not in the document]
JiriLojda: Oh, sure, I didn't notice 😄

src/modules/sync/syncModelRun.ts
JiriLojda: I meant to not fetch the entities at all when they are not included. Of course, we cannot filter them. To make sure the transformations work, we can maintain a map of dependencies between the entities and always fetch all the dependencies of all the included entities. That shouldn't be very complicated.

src/commands/syncModel/run/run.ts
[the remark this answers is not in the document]
IvanKiral: I think thats a creative solution, however, not sure whether it is not counter productive... I am starting to think that the best solution would be just to keep --entities parameters without --codenames and that's that.. If users want something more complex then they might use API

JiriLojda (approved):

src/modules/sync/sync.ts
```diff
@@ -1,71 +1,67 @@
...
 import { DiffModel } from "./types/diffModel.js";
 
-export const sync = async (client: ManagementClient, diff: DiffModel, logOptions: LogOptions) => {
+export const sync = async (
+  client: ManagementClient,
+  diff: DiffModel,
+  entities: ReadonlySet<SyncEntities>,
+  logOptions: LogOptions,
+) => {
+  entities as never;
```
JiriLojda, line 29: 😉

JiriLojda (changes requested):

src/modules/sync/sync/snippets.ts
```diff
@@ -79,6 +94,12 @@ export const addElementsIntoSnippetsWithoutReferences = async (
     ] as const
   );
 
+  if (addSnippetsOpsWithoutRefs.every(([, ops]) => !ops.length)) {
+    logInfo(logOptions, "standard", "No elements do add into content type snippets");
```
JiriLojda, line 98: typo

src/modules/sync/sync/webSpotlight.ts
```diff
@@ -1,14 +1,28 @@
...
+export const updateWebSpotlight = (
+  client: ManagementClient,
+  diffModel: WebSpotlightDiffModel,
+  logOptions: LogOptions,
+): Promise<unknown> =>
   match(diffModel)
-    .with({ change: "none" }, () => Promise.resolve())
-    .with({ change: "deactivate" }, () => client.deactivateWebSpotlight().toPromise())
+    .with({ change: "none" }, () => {
+      logInfo(logOptions, "standard", "No web spotlight to update");
```
JiriLojda, line 14: This sounds like the environment doesn't have web spotlight enabled. Something like "No web spotlight changes to perform" would be better.

src/modules/sync/sync/assetFolders.ts
```diff
@@ -15,12 +15,13 @@ export const syncAssetFolders = async (
   operations: DiffModel["assetFolders"],
   logOptions: LogOptions,
 ) => {
-  logInfo(logOptions, "standard", "Updating asset folders");
-
   if (!operations.length) {
+    logInfo(logOptions, "standard", "No asset folders to update");
```
JiriLojda, line 19: Just to avoid confusion with other entities that have `add/update/delete`, I would use a different wording here. e.g. "No asset folders updates"

src/modules/sync/sync/collections.ts
```diff
@@ -1,28 +1,45 @@
...
+  client: ManagementClient,
+  collections: DiffModel["collections"],
+  logOptions: LogOptions,
+) => {
   if (!collections.length) {
+    logInfo(logOptions, "standard", "No collections to update");
     return Promise.resolve();
   }
 
+  logInfo(logOptions, "standard", "Updating collections");
```
JiriLojda, line 18: To avoid confusion that there is no "Adding collections", I would instead use "Adding and updating collections". Similarly for the "No..." version.

JiriLojda (changes requested):

JiriLojda (approved):

PR #78, opened 2024-09-23
Devrel 1202/files validation

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/modules/sync/validation/elementSchemas.ts
+++ b/src/modules/sync/validation/elementSchemas.ts
@@ -1,155 +1,168 @@
+import { ContentTypeElements } from "@kontent-ai/management-sdk";
 import { z } from "zod";
 
-import { AddPropToObjectTuple, CombineTuples } from "../../../utils/types.js";
+import { AddPropToObjectTuple, IsFullEnum, IsSubset } from "../../../utils/types.js";
 import {
-  SyncAssetElement,
-  SyncCustomElement,
-  SyncDateTimeElement,
-  SyncGuidelinesElement,
-  SyncLinkedItemsElement,
-  SyncMultipleChoiceElement,
-  SyncNumberElement,
-  SyncRichTextElement,
+  SyncSnippetAssetElement,
+  SyncSnippetCustomElement,
+  SyncSnippetDateTimeElement,
+  SyncSnippetGuidelinesElement,
+  SyncSnippetLinkedItemsElement,
+  SyncSnippetMultipleChoiceElement,
+  SyncSnippetNumberElement,
+  SyncSnippetRichTextElement,
+  SyncSnippetTaxonomyElement,
+  SyncSnippetTextElement,
   SyncSubpagesElement,
-  SyncTaxonomyElement,
-  SyncTextElement,
   SyncTypeSnippetElement,
   SyncUrlSlugElement,
 } from "../types/syncModel.js";
 import { CodenameReferenceSchema } from "./commonSchemas.js";
 
-export const AssetElementDataSchema = z.strictObject({
-  name: z.string(),
-  codename: z.string(),
-  type: z.literal("asset"),
-  guidelines: z.string().optional(),
-  asset_count_limit: z
-    .strictObject({
-      value: z.number(),
-      condition: z.union([z.literal("at_most"), z.literal("exactly"), z.literal("at_least")]),
-    })
-    .optional(),
-  maximum_file_size: z.number().optional(),
-  allowed_file_types: z.union([z.literal("adjustable"), z.literal("any")]).optional(),
-  image_width_limit: z
-    .strictObject({
-      value: z.number(),
-      condition: z.union([z.literal("at_most"), z.literal("exactly"), z.literal("at_least")]),
-    })
-    .optional(),
-  image_height_limit: z
-    .strictObject({
-      value: z.number(),
-      condition: z.union([z.literal("at_most"), z.literal("exactly"), z.literal("at_least")]),
-    })
-    .optional(),
-  is_required: z.boolean().optional(),
-  is_non_localizable: z.boolean().optional(),
-  default: z
-    .strictObject({
-      global: z.strictObject({
-        value: z.array(CodenameReferenceSchema.extend({ external_id: z.string() })),
-      }),
-    })
-    .optional(),
-}) satisfies z.ZodType<SyncAssetElement>;
+type RequiredZodObject<T> = { [K in keyof T]-?: z.ZodType<T[K]> };
 
-export const CustomElementDataSchema = z.strictObject({
-  name: z.string(),
-  codename: z.string(),
-  type: z.literal("custom"),
-  source_url: z.string(),
-  json_parameters: z.string().optional(),
-  guidelines: z.string().optional(),
-  is_required: z.boolean().optional(),
-  is_non_localizable: z.boolean().optional(),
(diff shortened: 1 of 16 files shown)


src/modules/sync/validation/elementSchemas.ts
```diff
@@ -0,0 +1,323 @@
...
+  MultipleChoiceElementDataSchema,
+  RichTextElementDataSchema,
+  TaxonomyElementDataSchema,
+  TextElementDataSchema,
+] as const;
+
+const extendTuples = <Tuple1 extends ReadonlyArray<unknown>, Tuple2 extends ReadonlyArray<unknown>>(
+  t1: Tuple1,
+  t2: Tuple2,
+): CombineTuples<Tuple1, Tuple2> => [...t1, ...t2] as unknown as CombineTuples<Tuple1, Tuple2>;
```
JiriLojda, line 302: The casting is not necessary and in fact, the whole `extendTuples` is not needed. You can do the plain `[...SnippetElementsSchemas, SnipetElementDataSchema, ...] as const` and it works the same.

src/modules/sync/validation/elementSchemas.ts
```diff
@@ -0,0 +1,323 @@
...
+
+const TypeElementSchemas = extendTuples(
+  SnippetElementsSchemas,
+  [SnippetElementDataSchema, SubpagesElementDataSchema, UrlSlugElementDataSchema] as const,
+);
+
+export const TypeElementSchemasWithGroups = TypeElementSchemas
+  .map(schema =>
+    schema.extend({ content_group: z.strictObject({ codename: z.string() }) })
```
JiriLojda, line 311: Why don't you use the `ContentGroupSchema` here?

src/modules/sync/validation/elementSchemas.ts
```diff
@@ -0,0 +1,323 @@
...
+  ) as unknown as AddPropToObjectTuple<
+    typeof TypeElementSchemas,
+    typeof ContentGroupSchema
+  >;
+
+const ContentGroupSchema = z.strictObject({
+  content_group: z.strictObject({ codename: z.string() }),
```
JiriLojda, line 318: Why don't you use the `CodenameReferenceSchema` from `commonSchemas.ts`?

src/modules/sync/validation/entitySchema.ts
```diff
@@ -0,0 +1,164 @@
...
+  codename: z.string(),
+  elements: z.array(SnippetElementsSchemasUnion),
+});
+
+const ContentGroupSchema = z.strictObject({ codename: z.string(), name: z.string() });
+
+export const TypeSchema: z.ZodType<
+  ContentTypeSyncModel,
+  z.ZodTypeDef,
+  { content_groups: ReadonlyArray<Replace<ContentTypeContracts.IContentTypeGroup, { codename: string }>> }
```
JiriLojda, line 46:
```suggestion
  Pick<ContentTypeSyncModel, "content_groups">
```

src/modules/sync/validation/entitySchema.ts
```diff
@@ -0,0 +1,164 @@
...
+      name: z.string(),
+      codename: z.string(),
+      is_active: z.boolean(),
+      is_default: z.literal(false),
+      fallback_language: CodenameReferenceSchema,
+    },
+  ),
+]);
+
+const WorkflowColorSchema = z.union([
```
JiriLojda, line 114: Don't forget to use explicit types for the schemas as it is easy to make a mistake here, especially in this union. If you forget a a color, the TS won't complain even when you use an explicit type for the constant. The best solution is probably to have an explicitly typed object with keys being the colors and values being the appropriate schema and then generate the union from the object with `Object.values`. This way TS would check that you don't forget a color.

src/modules/sync/validation/entitySchema.ts
```diff
@@ -0,0 +1,164 @@
...
+  z.literal("orange"),
+  z.literal("brown"),
+]);
+
+const WorkflowPublishedStepSchema = z.strictObject({
+  name: z.string(),
+  codename: z.string(),
+  create_new_version_role_ids: z.array(z.string()).length(0),
+  unpublish_role_ids: z.array(z.string()).length(0),
+});
```
JiriLojda, line 138: Use `satisfies` in all the `strictObject` schemas with all the properties set to required as that's the best way to ensure we don't forget even an optional property.

tests/unit/syncModel/validation/languageSchema.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      codename: null,
+      is_active: null,
+      is_default: true,
+    };
+
+    const result = LanguageSchema.safeParse(invalidData);
+    expect(result.success).toBe(false);
+  });
+
+  test("Invalid object that is empty", () => {
```
JiriLojda, line 129: We use `describe/it` functions in this project. Although `test` and `it` behave the same, I would stick to only one of them.

tests/unit/syncModel/validation/languageSchema.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      fallback_language: {
+        codename: "en",
+      },
+    };
+
+    const result = LanguageSchema.safeParse(invalidData);
+    expect(result.success).toBe(false);
+  });
+
+  test("Invalid object with is_default false and missing fallback_language", () => {
```
JiriLojda, line 48: Although I like the enthusiasm with tests, I would rely on TS forcing us to create the schemas properly and then `zod` to behave properly with the right schemas. You didn't write tests on all the schemas anyway. I think that one valid and one invalid test per schema is more than enough and instead we can cover all the schemas.

src/modules/sync/validation/entitySchema.ts
JiriLojda: What is the difference between `z.enum` and `z.union` filled with `z.literal`. I noticed you used both in different places. Can we use the `z.enum` here as well? It seems to be simpler.

JiriLojda (changes requested):

src/utils/types.ts
```diff
@@ -67,3 +67,71 @@ export type AnyOnePropertyOf<Obj extends object> = [keyof Obj, keyof Obj] extend
...
+ * type Invalid = CombineTuples<[], ReadonlyArray<number>>;
+ * // Result: "never"
+ *
+ * @description
+ * - If `Tuple1` or `Tuple2` extends `ReadonlyArray<unknown>`, meaning they are not fixed-length tuples,
+ *   the resulting type is `"never"`.
+ */
+
+export type CombineTuples<Tuple1 extends ReadonlyArray<unknown>, Tuple2 extends ReadonlyArray<unknown>> =
+  ReadonlyArray<unknown> extends Tuple1 ? "never" : ReadonlyArray<unknown> extends Tuple2 ? "never"
```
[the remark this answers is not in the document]
IvanKiral: after doing other comments the type is not necessary so i removed it

src/modules/sync/validation/elementSchemas.ts
IvanKiral: because sadly extend does take raw shape

IvanKiral: i actually used ContentGroupSchema only for type

JiriLojda (approved):

PR #96, opened 2024-11-04
Add migrations improvements

JiriLojda:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/commands/migrations/add/add.ts
+++ b/src/commands/migrations/add/add.ts
@@ -1,5 +1,5 @@
 import { logError, LogOptions } from "../../../log.js";
-import { addMigration } from "../../../modules/migrations/add.js";
+import { addMigration, AddMigrationParams } from "../../../modules/migrations/add.js";
 import { RegisterCommand } from "../../../types/yargs.js";
 import { simplifyErrors } from "../../../utils/error.js";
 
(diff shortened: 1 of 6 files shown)


src/modules/migrations/add.ts
```diff
@@ -59,3 +59,16 @@ export const addMigration = async (params: AddMigrationParams) => {
     `Migration ${migrationName} has been created sucessfully in ${chalk.blue(saveMigrationPath)}`,
   );
 };
+
+const padWithLeadingZeros = (order: number, padWithLeadingZeros: number | undefined) =>
```
IvanKiral, line 63: Nipick: Is the parameter called the same as the function? I am surprised that the compiler does not yell :D (how does it know that you don't want to call that function recursively)... I would rename the parameter

IvanKiral (changes requested):

src/commands/migrations/add/add.ts
```diff
@@ -34,7 +36,19 @@ export const register: RegisterCommand = yargs =>
...
+        })
+        .option("order", {
+          alias: "o",
+          describe: "Sets the order property to the provided value.",
+          type: "number",
+          conflicts: ["timestamp"],
+        })
+        .option("padWithLeadingZeros", {
+          describe: "Specifies the number of leading zeros for the order property.",
+          type: "number",
```
IvanKiral, line 49: I think you meant for the name of the migration

IvanKiral (approved):

PR #102, opened 2024-11-20
fix not showing diff when no entities are specified

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/modules/sync/sync/workflows.ts
+++ b/src/modules/sync/sync/workflows.ts
@@ -17,18 +17,24 @@ export const syncWorkflows = async (
     logInfo(logOptions, "standard", "No workflows to add");
   }
 
-  if ([...operations.updated].flatMap(([, arr]) => arr).length) {
+  const workflowCodenamesToUpdate = [...operations.updated.entries()]
+    .filter(([, arr]) => arr.length)
+    .map(([codename]) => codename);
+
+  if (workflowCodenamesToUpdate.length) {
     logInfo(logOptions, "standard", "Updating workflows");
 
     await serially(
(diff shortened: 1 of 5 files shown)


src/modules/sync/sync/workflows.ts
```diff
@@ -21,14 +21,16 @@ export const syncWorkflows = async (
     logInfo(logOptions, "standard", "Updating workflows");
 
     await serially(
-      [...operations.updated.keys()].map(codename => () =>
-        modifyWorkflow(
+      [...operations.updated.entries()].filter(([, arr]) => arr.length).map(([key]) => key).map(codename => () => {
```
pokornyd, line 24: this feels kinda convoluted, don't you think? I see we check whether there are any update operations on line 20, then we perform similar thing on line 24. how about extracting the codenames of workflows to be updated to a separate array, then check its length and if it's >1, perform the actions, otherwise log the "No workflows to update" info? something like this:

```ts
const workflowCodenamesToUpdate = [...operations.updated.entries()].filter(([, arr]) => arr.length).map(([key]) => key);

  if (workflowCodenamesToUpdate.length) {
    logInfo(logOptions, "standard", "Updating workflows");

    await serially(
      workflowCodenamesToUpdate.map(codename => () => { } // rest of the logic
```

pokornyd (changes requested):

src/modules/sync/sync/workflows.ts
IvanKiral: good catch, updated :)

pokornyd (approved):

PR #108, opened 2024-12-06
add headers to webhook data

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/utils/files.ts
+++ b/src/utils/files.ts
@@ -1,3 +1,5 @@
+import { padWithLeadingZeros } from "./number.js";
+
 export enum DateLevel {
   Year = 0,
   Month = 1,
@@ -8,11 +10,18 @@ export enum DateLevel {
 }
 
 export const serializeDateForFileName = (date: Date, level: DateLevel) =>
-  [
-    date.getUTCFullYear(),
-    ...level >= DateLevel.Month ? [("0" + (date.getUTCMonth() + 1)).slice(-2)] : [],
-    ...level >= DateLevel.Day ? [("0" + date.getUTCDate()).slice(-2)] : [],
-    ...level >= DateLevel.Hour ? [("0" + date.getUTCHours()).slice(-2)] : [],
-    ...level >= DateLevel.Minute ? [("0" + date.getUTCMinutes()).slice(-2)] : [],
-    ...level >= DateLevel.Second ? [("0" + date.getUTCSeconds()).slice(-2)] : [],
-  ].join("-");
+  createDateParts(date)
+    .slice(0, level + 1)
+    .join("-");
+
+const createDateParts = (date: Date) =>
+  Object.values(
+    {
+      [DateLevel.Year]: date.getUTCFullYear().toString(),
+      [DateLevel.Month]: padWithLeadingZeros(date.getUTCMonth() + 1, 2),
+      [DateLevel.Day]: padWithLeadingZeros(date.getUTCDate(), 2),
+      [DateLevel.Hour]: padWithLeadingZeros(date.getUTCHours(), 2),
+      [DateLevel.Minute]: padWithLeadingZeros(date.getUTCMinutes(), 2),
+      [DateLevel.Second]: padWithLeadingZeros(date.getUTCSeconds(), 2),
+    } as const satisfies Record<DateLevel, string>,
+  );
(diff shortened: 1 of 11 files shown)


src/utils/files.ts
```diff
@@ -1,4 +1,6 @@
 export const serializeDateForFileName = (date: Date) =>
-  `${date.getUTCFullYear()}-${
-    date.getUTCMonth() + 1
-  }-${date.getUTCDate()}-${date.getUTCHours()}-${date.getUTCMinutes()}`;
+  `${date.getUTCFullYear()}-`
```
JiriLojda, line 2: There is a `formatDateForFileName` function in `src/modules/migrations/utils/migrationUtils.ts` that has the same implementation. Remove it and replace its calls to this function instead.

JiriLojda: Oh, I noticed it also has seconds in the result. Maybe we can parametrize this?

JiriLojda (changes requested):

src/utils/files.ts
JiriLojda: I didn't have anything in mind. 😄 But it's true that the single boolean parameter is rather random. Since we don't want formats like YYYY-DD, we can instead have a parameter specifying the maximum detail (e.g. "year" -> YYYY, "minute" -> YYYY-DD-hh-mm").

JiriLojda: I pushed a suggested refactor of the function. Feel free to remove/squash it.

src/utils/files.ts
```diff
@@ -1,4 +1,25 @@
...
+
+export const serializeDateForFileName = (date: Date, level: DateLevel) =>
+  createDateParts(date)
+    .slice(0, level + 1)
+    .join("-");
+
+const createDateParts = (date: Date) =>
+  Object.values(
+    {
+      [DateLevel.Year]: date.getUTCFullYear().toString(),
```
JiriLojda, line 18: It's very unlikely, but if the date is smaller than 4 digits, then the number wouldn't have 4 digits. Is it ok?

src/utils/files.ts
```diff
@@ -1,4 +1,25 @@
...
+      [DateLevel.Month]: ("0" + (date.getUTCMonth() + 1)).slice(-2),
+      [DateLevel.Day]: ("0" + date.getUTCDate()).slice(-2),
```
JiriLojda, line 20: We have a function called `padWithLeadingZeros` in the `src/modules/migrations/add.ts` module that does exactly this. It would be nice to reuse it.

JiriLojda (changes requested):

tests/unit/utils/number.test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, it } from "vitest";
+
+import { padWithLeadingZeros } from "../../../src/utils/number.js";
+
+describe("padWithLeadingZeros", () => {
+  it("should pad the number with leading zeros when numberOfZeros is greater than the length of the number", () => {
+    expect(padWithLeadingZeros(5, 3)).toBe("005");
+    expect(padWithLeadingZeros(42, 5)).toBe("00042");
+    expect(padWithLeadingZeros(0, 4)).toBe("0000");
```
JiriLojda, line 9: Prefer parametrized tests over one test with multiple expects each for a different case.

tests/unit/utils/files.test.ts
```diff
@@ -0,0 +1,20 @@
...
+describe("serializeDateForFileName", () => {
+  it.each([
+    ["2025", DateLevel.Year],
+    ["2025-02", DateLevel.Month],
+    ["2025-02-07", DateLevel.Day],
+    ["2025-02-07-15", DateLevel.Hour],
+    ["2025-02-07-15-08", DateLevel.Minute],
+    ["2025-02-07-15-08-40", DateLevel.Second],
+  ])(
+    "correctly serializes datetime",
```
JiriLojda, line 14: Include the case data in the test-case name so the cases are distinguishable. Now it is not possible to know which case failed. You can use there printf formatting. https://vitest.dev/api/#test-each

JiriLojda (changes requested):

JiriLojda (approved):
Lets go

PR #110, opened 2024-12-13
Fix typos

andgdk:
### Motivation

Correct spelling in readme, commands, symbols, parameters, and tests.
The second commit addresses a build error, as type inference was missing.

### Checklist

- [x] Code follows coding conventions held in this repo
- [ ] Tests are passing
- [x] Temporary settings (e.g. variables used during development and testing) have been reverted to defaults

### DIFF
--- a/src/commands/environment/backupRestore/README.md
+++ b/src/commands/environment/backupRestore/README.md
@@ -1,6 +1,6 @@
 # Environment Backup & Restore
 
-The `environment backup` and `enviroment restore` environment commands let you easily backup and then restore your Kontent.ai environment.
+The `environment backup` and `environment restore` environment commands let you easily backup and then restore your Kontent.ai environment.
 
 ## Including and Excluding Entities in Environment commands
 
--- a/src/commands/migrateContent/run/run.ts
+++ b/src/commands/migrateContent/run/run.ts
@@ -33,20 +33,20 @@ export const register: RegisterCommand = yargs =>
         })
         .option("sourceEnvironmentId", {
           type: "string",
-          describe: "Id of Kontent.ai environmnent containing source content.",
+          describe: "Id of Kontent.ai environment containing source content.",
           alias: "s",
           implies: ["sourceApiKey"],
         })
         .option("sourceApiKey", {
           type: "string",
-          describe: "Management Api key of Kontent.ai environmnent containing source content.",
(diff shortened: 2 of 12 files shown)


src/modules/sync/diff/workflow.ts
```diff
@@ -16,7 +16,7 @@ const workflowStepHandler: Handler<WorkflowStepSyncModel> = makeObjectHandler({
   color: baseHandler,
   transitions_to: makeArrayHandler(
     t => t.step.codename,
-    makeObjectHandler({ step: makeLeafObjectHandler({ codename: (source, target) => source === target }) }),
+    makeObjectHandler({ step: makeLeafObjectHandler({ codename: (source: any, target: any) => source === target }) }),
```
IvanKiral, line 19: why is this needed? we don't do `any` here unless completely necessary

src/modules/sync/diff/language.ts
```diff
@@ -14,7 +14,7 @@ export const languageHandler: Handler<LanguageSyncModel> = makeObjectHandler({
   is_active: baseHandler,
   is_default: constantHandler,
   fallback_language: optionalHandler(makeLeafObjectHandler({
-    codename: (source, target) => source === target,
+    codename: (source: any, target: any) => source === target,
```
IvanKiral, line 17: same here as other comment

IvanKiral (changes requested):
Thanks for finding all the typos! It seems I would need to have Grammarly in my IDE. Left the comments for type interference tho, we don't use any in this codebase unless necessary. Can you please explain why it is needed?

andgdk:
> ### Motivation
> [...] The second commit addresses a build error, as type inference was missing.

Not sure why this was. I will investigate further and undo those changes.

src/modules/sync/diff/workflow.ts
andgdk: Understand. I can't reproduce and happyly reverted.

src/modules/sync/diff/language.ts
andgdk: Reverted

andgdk:
@IvanKiral Made the requested change. Thanks for taking the time to review. I end up with typos myself even with tooling - but yes, can be a life saver :)

IvanKiral (approved):

IvanKiral:
Thanks for fixing :)

PR #125, opened 2025-01-29
makeLeafObjectHandler

### LINKED ISSUE #124: False positives on diff
### Brief bug description

Getting items on snapshot diff reports that don't seem to differ

### Repro steps

1. diff 2 similar environments -> no difference
2. snapshot one of the environments
3. diff 1 environment to snapshot
4. diff report shows changes, but they don't seem to be different

### Expected behavior

no difference shown in both diffs

### Test environment

 - Platform/OS: MacOS 15.3

### Additional context

I would post a screenshot, but it's validation regex in content types so I don't want to risk a security issue

IvanKiral:
### Motivation

Which issue does this fix? Fixes #124 

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/modules/sync/diff/combinators.ts
+++ b/src/modules/sync/diff/combinators.ts
@@ -52,11 +52,16 @@ export const makeLeafObjectHandler = <Entity extends object>(
   transformBeforeReplace: (v: Entity) => unknown = x => x,
 ): Handler<Entity> =>
 (sourceValue, targetValue) => {
-  const shouldReplace = zip(
-    Object.keys(sourceValue) as (keyof typeof customComparers)[],
-    zip(Object.values(sourceValue), Object.values(targetValue)),
-  )
-    .some(([key, [source, target]]) => !(customComparers[key]?.(source, target) ?? source === target));
+  const keys = new Set([
+    ...Object.keys(sourceValue),
+    ...Object.keys(targetValue),
+    ...Object.keys(customComparers),
+  ]) as Set<keyof Entity>;
+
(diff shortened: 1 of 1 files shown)


src/modules/sync/diff/combinators.ts
```diff
@@ -52,11 +52,16 @@ export const makeLeafObjectHandler = <Entity extends object>(
...
-    .some(([key, [source, target]]) => !(customComparers[key]?.(source, target) ?? source === target));
+  const keys = new Set([
+    ...Object.keys(sourceValue),
+    ...Object.keys(targetValue),
+    ...Object.keys(customComparers),
+  ]) as Set<keyof Entity>;
+
+  const shouldReplace = Array.from(keys)
+    .some(key =>
+      !(customComparers[key]?.(sourceValue[key], targetValue[key]) ?? sourceValue[key] === targetValue[key])
```
JiriLojda, line 63: I know that it was there before, but is this part `sourceValue[key] === targetValue[key]` in the expression correct? I would think that we want to replace the object if any of the properties are different, meaning this expression instead `sourceValue[key] !== targetValue[key]`. Am I missing something? Can you write a test case for this scenario if it is indeed a bug?

JiriLojda (changes requested):

src/modules/sync/diff/combinators.ts
JiriLojda: Oh, I didn't notice the negation is around the whole expression. Everything is alright then.

JiriLojda (approved):

PR #128, opened 2025-02-04
Formatted select files so my actual changes can be clearer in a future PR

bbuurstra-bzs:
No actual code changes. Just default prettier formatting applied to files that have been edited in my actual branch with my real logic updates


bbuurstra-bzs:
Didn't mean to create a PR back into the main repo 🤦 Moving too fast... Disregard!

PR #135, opened 2025-04-02
Bump dependencies

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -1,41 +1,41 @@
...
+  return [];
 };
 
-const customResolveDomTextNode: ResolveDomTextNodeType = () => null;
+const resolveItemIdsDomHtmlNode = (node: PortableTextObject): ReadonlyArray<string> => {
+  if (node._type === "componentOrItem") {
+    return node.dataType === "item" ? [node.componentOrItem._ref] : [];
+  }
+  if (node._type === "block") {
+    return node.markDefs?.filter(a => a._type === "contentItemLink").map(a => (a as any).contentItemLink._ref) ?? [];
```
JiriLojda, line 30: Can you use this type instead of any? 
image
Also, according to the type the `a.contentItemLink._ref` is an object, not a string. Does the code work?

src/modules/sync/utils/contentTypeHelpers.ts
```diff
@@ -1,41 +1,41 @@
...
-    default: {
-      return node.children.map(traverse);
-    }
+const resolveAssetIdsDomHtmlNode = (node: PortableTextObject): ReadonlyArray<string> => {
+  if (node._type === "image") {
+    return node.asset.referenceType === "id" ? [node.asset._ref] : [];
+  }
+
+  if (node._type === "block") {
+    return node.markDefs?.filter(a => a._type === "link" && "data-asset-id" in a).map(a => (a as any)["data-asset-id"])
```
JiriLojda, line 14: Similarly can you use a different type here other than any?

JiriLojda (changes requested):

src/modules/sync/utils/contentTypeHelpers.ts
IvanKiral: no contentItemLink is reference ._ref is a string :D however casting to this object sounds better, thanks :)

also the `getRequiredIds` that utilizes this functions are tested, so yeah it should be working :)

src/modules/sync/utils/contentTypeHelpers.ts
IvanKiral: Sadly I think that for asset link the Type is missing :/

src/modules/sync/utils/contentTypeHelpers.ts
JiriLojda: Oh, you are right

JiriLojda (approved):

PR #153, opened 2025-08-01
Migrate Content & Sync Run Suppress message fix

carlo-kontent:
### Motivation

Which issue does this fix? 
New-found issue. 

Migrate Content & Sync Run suppress message flag (--sw) was incorrect (eg. suppress this message with --sw parameter) and did not match the suppress command flags for each command (eg. skipConfirmation). Migrate Content & Sync Run suppress flags and messages have been updated to match the Clean command (eg. --s)

### Checklist

- [ ] Tests are passing

### How to test

When using the Migrate Content & Sync Run, the suppress flag has been updated from m--sw to --s. An alias for each command has also been created.

### DIFF
--- a/src/commands/migrateContent/run/run.ts
+++ b/src/commands/migrateContent/run/run.ts
@@ -105,6 +105,7 @@ export const register: RegisterCommand = (yargs) =>
         .option("skipConfirmation", {
           type: "boolean",
           describe: "Skip confirmation message.",
+          alias: "s",
         })
         .option("kontentUrl", {
           type: "string",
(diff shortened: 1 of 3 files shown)


src/commands/migrateContent/run/run.ts
```diff
@@ -105,6 +105,7 @@ export const register: RegisterCommand = (yargs) =>
         .option("skipConfirmation", {
           type: "boolean",
           describe: "Skip confirmation message.",
+          alias: "s",
```
IvanKiral, line 108: we use "s" as alias for sourceEnvironmentId

src/commands/sync/run/run.ts
```diff
@@ -70,6 +70,7 @@ export const register: RegisterCommand = (yargs) =>
         .option("skipConfirmation", {
           type: "boolean",
           describe: "Skip confirmation message.",
+          alias: "s",
```
IvanKiral, line 73: we use "s" as alias for sourceEnvironmentId

IvanKiral (changes requested):

PR #160, opened 2025-09-30
fix importing language variants to be the default first

### LINKED ISSUE #159: Restore fails when publishing languageVariants with non-localizable elements
When restoring an environment using `@kontent-ai/data-ops`, the process fails with the following error if content items contain non-localizable elements:

Error:
{"message":"Bad Request – Publishing item failed because some non-localizable elements need to be published in the default variant.","errorCode":4040031}

It looks like the tool attempts to publish non-default language variants before publishing the default language variant. Since non-localizable fields must be published in the default variant first, the restore fails.

Steps to reproduce:
1. Create a content type with a required non-localizable element (e.g., asset, taxonomy).
2. Publish items with multiple language variants.
3. Export environment.
4. Try to restore into a clean environment.
5. Restore fails at the `languageVariants` step with the error above.

Expected behavior:
The tool should first publish the default language variant, then proceed with other language variants.

Workaround:
Running restore with `--skipPublish` works, but requires manual publishing afterwards.

Tool version: @kontent-ai/data-ops @kontent-ai/data-ops (latest as of 2025-09-29)

Node.js version: 20.19.4

OS: Mac os

IvanKiral:
### Motivation

Which issue does this fix? Fixes #159 

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@kontent-ai/data-ops",
-  "version": "2.3.13",
+  "version": "2.4.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@kontent-ai/data-ops",
-      "version": "2.3.13",
+      "version": "2.4.0",
       "license": "MIT",
       "dependencies": {
         "@kontent-ai/core-sdk": "^10.10.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kontent-ai/data-ops",
-  "version": "2.3.13",
+  "version": "2.4.0",
   "description": "",
   "type": "module",
   "scripts": {


src/modules/backupRestore/backupRestoreEntities/entities/languageVariants.ts
```diff
@@ -44,7 +44,23 @@ export const languageVariantsEntity = {
...
-    await serially(fileVariants.map(createImportVariant(client, context, logOptions)));
+    await serially(
+      fileVariants
+        .toSorted((a, b) => {
+          if (a.item.id === b.item.id) {
+            if (a.language.id === emptyId) {
+              return -1;
+            }
+
+            if (b.language.id === emptyId) {
```
JiriLojda: I don't think this can ever happen. Both of these `if`s are wrapped in `aId === bId` and the first `if` returns when `aId` === `emptyId` and if that's false `bId` cannot be empty as well, because in this branch you can only be when `aId` === `bId`. My guess is that the top `if` should have `!==` instead.

JiriLojda: It's ease to make mistake in these `if`s. If we have `ts-pattern` installed here, I'd suggest using it here. Something like
```
match([a.item.id, b.item.id])
  .with([emptyId, emptyId], () => 0)
  .with([emptyId, P._], () => -1)
  .with(P._, emptyId], () => 1)
  .otherwise(() => 0)
```
could work.

JiriLojda (changes requested):

JiriLojda (approved):

PR #166, opened 2025-12-02
update advancedDiff to react

### LINKED ISSUE #171: sync diff creates file instead of subdirectory
### Brief bug description

If `sync diff` is run within a directory that contains a dot (.) as part of its naming hierarchy, then the `outPath` parameter (given that it simply specifies a directory name) is used to create a file with the output instead of a subdirectory containing a file named `diff_<current UTC dateTime>.html`.

### Repro steps

1. Open a command prompt in a directory that contains a dot in its naming hierarchy, e.g. `c:\dev\my.project\tools`.
2. Run `npx @kontent-ai/data-ops@latest sync diff --outPath Output` supplying the other necessary arguments.

### Actual behavior

A file called `Output` will be created. Renaming it to `Output.html` and opening it in a browser reveals it to be the diff report file.

### Expected behaviour

A directory called `Output` should be created with a file called diff_<current UTC dateTime>.html.

### Test environment

 - Platform/OS: Windows 11
 - Browser: N/A
 - Version: Latest

### Additional context

This is the diagnosis of the issue from Claude Sonnet:

The bug is in resolveOutputPath in fileUtils.ts:8-14:

```
const hasExtension = path.resolve(outputPath).includes(".");
```

path.resolve("Output") produces the full absolute path including the current working directory. If that directory contains a . anywhere in it (e.g. C:\Users\john.doe\projects), then hasExtension becomes true even though "Output" has no extension — so the code treats it as a file path instead of a directory.

The fix is to check only the provided outputPath itself (or just its basename), not the fully resolved path:

Edited fileUtils.ts
Before:
```
export const resolveOutputPath = (outputPath: string) => {
  const hasExtension = path.resolve(outputPath).includes(".");
```
After:
```
export const resolveOutputPath = (outputPath: string) => {
  const hasExtension = path.extname(outputPath) !== "";
```

path.extname() checks only the extension of the filename component of the path, so a directory name like Output correctly returns "" regardless of what dots appear in the surrounding directory path.

### Screenshots

Image

IvanKiral:
### Motivation

Which issue does this fix? Fixes #171 

Apart from fixing the mentioned issue, this PR changes custom template system for html to react using `renderToString`.

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kontent-ai/data-ops",
-  "version": "2.5.0",
+  "version": "2.6.0",
   "description": "",
   "type": "module",
   "scripts": {


src/modules/sync/advancedDiff/components/entities/LanguagesSection.tsx
```diff
@@ -0,0 +1,34 @@
...
+  languages: DiffObject<
+    RequiredCodename<LanguageModels.IAddLanguageData> & Readonly<{ is_default: boolean }>
+  >;
+}>;
+
+export const LanguagesSection = ({ languages }: LanguagesSectionProps) => (
+  <DiffObjectSection
+    id="languages"
+    title="Languages"
+    noChangesMessage="No changes to languages"
```
JiriLojda: Other entities have dot at the end, add it here too.

src/modules/sync/advancedDiff/utils/valueHelpers.tsx
```diff
@@ -0,0 +1,151 @@
...
+  isExternalIdReference,
+  isExternalIdReferenceArray,
+  isMaximumTextLength,
+  isObjectReference,
+  isObjectReferenceArray,
+  isValidationRegex,
+} from "../../../../utils/typeguards.js";
+import { renderRichTextValue } from "./richTextResolvers.js";
+
+const getValueOrIdentifierInternal = (value: unknown): ReactNode | ReadonlyArray<ReactNode> => {
```
JiriLojda: I don't like this function, but have no idea how to make it better. :D

PR #167, opened 2025-12-25
Don't check file existence for entities not part of sync run

JiriLojda:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kontent-ai/data-ops",
-  "version": "2.4.0",
+  "version": "2.4.1",
   "description": "",
   "type": "module",
   "scripts": {


.github/workflows/build.yml
```diff
@@ -10,11 +10,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
```
IvanKiral: please updat to v6 for checkoutg and check version of others actions

.github/workflows/release.yml
```diff
@@ -9,20 +9,22 @@ jobs:
...
-      - run: npm run build
-      
-      - run: npm publish --access=public
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm build
+
+      - run: pnpm publish --access=public --no-git-checks
         if: ${{!github.event.release.prerelease}}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_API_KEY }}
```
IvanKiral: please remove node_auth_token  ( I tihnk its only for release) and change it to oidc

src/modules/sync/validation.ts
```diff
@@ -6,24 +6,25 @@ import chalk from "chalk";
...
-} from "./constants/filename.js";
+import { entityToFilename, type SyncEntityName } from "./constants/entities.js";
 import type { ElementsTypes } from "./types/contractModels.js";
 import type { DiffModel } from "./types/diffModel.js";
 import { getTargetCodename, type PatchOperation } from "./types/patchOperation.js";
 
-export const validateSyncModelFolder = async (folderPath: string) => {
+export const validateSyncModelFolder = async (
+  folderPath: string,
+  entities: ReadonlySet<SyncEntityName>,
```
IvanKiral: why is this set?

IvanKiral (changes requested):

src/modules/sync/validation.ts
JiriLojda: Conceptually, I have to agree with Claude that it makes sense for this to be a readonly set. Having said that, I do agree with you that it is unnecessary complication as we never have the data in this structure so I refactored the function to accept an array of entity names instead.

PR #177, opened 2026-05-13
Fix spaces validation

IvanKiral:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/backupRestore/backupRestoreEntities/entities/spaces.ts
```diff
@@ -37,16 +61,20 @@ export const spacesEntity = {
...
               ? {
-                  id: getRequired(
-                    context.contentItemContextByOldIds,
-                    importSpace.web_spotlight_root_item.id ?? "missing-ws-root-id",
-                    "item",
-                  ).selfId,
+                  root_item: {
+                    id: getRequired(
+                      context.contentItemContextByOldIds,
+                      importRootItem.id ?? "missing-ws-root-id",
```
JiriLojda: nit: No longer "ws-root", right?

src/modules/backupRestore/backupRestoreEntities/entities/spaces.ts
```diff
@@ -1,19 +1,35 @@
-import type { SpaceContracts } from "@kontent-ai/management-sdk";
+import type { SharedContracts, SpaceContracts, SpaceModels } from "@kontent-ai/management-sdk";
 import chalk from "chalk";
 
 import { logInfo } from "../../../../log.js";
 import { serially } from "../../../../utils/requests.js";
 import { getRequired } from "../../utils/utils.js";
 import type { EntityDefinition } from "../entityDefinition.js";
 
+// SDK does not yet expose `root_item` on ISpaceContract; MAPI returns it during
```
JiriLojda: Is the SDK still outdated? 😄  If so, the changes we will need to do in the future once the SDK updates are scattered across the codebase. Will we be able to locate and fix them all easily? All the custom types and casts that ts won't find automaticaly?

src/modules/backupRestore/backupRestoreEntities/entities/spaces.ts
```diff
@@ -25,6 +41,14 @@ export const spacesEntity = {
           `Importing: space ${importSpace.id} (${chalk.yellow(importSpace.name)})`,
         );
 
+        // Older backups carry the legacy `web_spotlight_root_item`; newer ones carry `root_item`.
```
JiriLojda: Do we need to handle old backups here? Shouldn't they be instead handled (transformed) when loading from file?

src/modules/sync/modelTransfomers/spaceTransformers.ts
```diff
@@ -1,29 +1,41 @@
...
 export const transformSpacesModel = (
   environmentModel: EnvironmentModel,
 ): ReadonlyArray<SpaceSyncModel> =>
-  environmentModel.spaces.map((space) => ({
-    ...omit(space, ["id"]),
-    web_spotlight_root_item: space.web_spotlight_root_item
-      ? {
+  environmentModel.spaces.map((space) => {
+    const rootItem =
+      (space as typeof space & SpaceWithRootItem).root_item ?? space.web_spotlight_root_item;
```
JiriLojda: Is the fallback needed here? I would expect the MAPI now returns both properties defined and we can expect the either both or none of them is, am I correct?

JiriLojda (changes requested):

src/modules/backupRestore/backupRestoreEntities/entities/spaces.ts
IvanKiral: it is outdated, because its not even documented yet :D Well I hope we will use llms for that :)

IvanKiral: also trying to unified that. those llms really love redundancy

JiriLojda (approved):

PR #178, opened 2026-05-13
Add live preview

IvanKiral:
### Motivation

based on branch #177

### Checklist

- [ ] Tests are passing

### How to test


src/commands/sync/run/run.ts
```diff
@@ -134,9 +139,24 @@ const resolveParams = (params: SyncModelRunCliParams): SyncRunParams => {
...
-    ...entities.filter((a) => a !== "webSpotlight").map((e) => [e, () => true]),
-    ...(entities.includes("webSpotlight") ? [["webSpotlight", true]] : []),
+    ...entities.filter((a) => a !== "livePreview").map((e) => [e, () => true]),
+    ...(entities.includes("livePreview") ? [["livePreview", true]] : []),
   ] as const;
 
   return Object.fromEntries(filterEntries);
 };
+
+const normalizeWebSpotlightAlias = (
```
JiriLojda: This function is duplicated here and in the snapshot commands. On top of that it is not used in the diff command so when user types webSpotlight entity there it is just ignored from the diff (both cli diff and advanced diff).

src/public.ts
```diff
@@ -45,9 +45,9 @@ export {
   ContentTypeSnippetsSyncModel,
   ContentTypeSyncModel,
   LanguageSyncModel,
+  LivePreviewSyncModel,
   SpaceSyncModel,
   SyncSnippetElement,
   SyncTypeElement,
   TaxonomySyncModel,
-  WebSpotlightSyncModel,
```
JiriLojda, line 52: This is a breaking change.

src/modules/backupRestore/backupRestoreEntities/entityDefinition.ts
```diff
@@ -41,6 +41,9 @@ type EntityRestoreParams<T> = {
 
 export type EntityRestoreDefinition<T> = Readonly<{
   name: string;
+  // Optional name of a legacy ZIP entry to fall back to when `${name}.json` is absent.
+  // Used for backward compatibility when an entity's file name changes between versions.
+  legacyName?: string;
```
JiriLojda: Can we make this more future-proofed by having `legacyNames?: ReadonlyArray<string>` instead? It is an easy change and will work with however renames we will need in the future. :)

tests/integration/backupRestore/utils/envData.ts
```diff
@@ -211,7 +221,17 @@ export const loadAllEnvDataFromZip = async (fileName: string): Promise<AllEnvDat
     assetFolders: await loadFile(zip, "assetFolders.json"),
     assets: await loadFile(zip, "assets.json"),
     webhooks: await loadFile(zip, "webhooks.json"),
-    webSpotlight: await loadFile(zip, "webSpotlight.json"),
+    // New backups produce `livePreview.json`; older backups produced `webSpotlight.json`
+    // with `{enabled, root_type}` and are normalized here for backward-compat tests.
```
JiriLojda: Do we need to put this rather complicated logic here just because one test suite? Can that test suite handle that itself and rest of the tests assume we have up-to-date .zip file for tests?

src/modules/sync/types/syncModel.ts
```diff
@@ -134,10 +133,7 @@ export type ContentTypeSyncModel = Replace<
   }>
 >;
 
-export type WebSpotlightSyncModel = Replace<
-  WebSpotlightContracts.IWebSpotlightStatus,
-  { root_type: Readonly<{ codename: string }> | null }
->;
+export type LivePreviewSyncModel = Readonly<{ status: string }>;
```
JiriLojda: This can only be `"enabled" | "disabled"` or something like this, right?

JiriLojda (changes requested):

src/public.ts
IvanKiral: Let's keep it and use @dperecated tag then.

src/commands/sync/diff/diff.ts
```diff
@@ -115,7 +116,10 @@ const syncDiffCli = async (params: syncDiffCliParams) => {
 };
 
 const resolveParams = (params: syncDiffCliParams): SyncDiffParamsInternal => {
-  const updatedParams = { ...params, entities: params.entities ?? syncEntityChoices };
+  const updatedParams = {
+    ...params,
+    entities: normalizeWebSpotlightAlias(params.entities ?? syncEntityChoices, params),
```
JiriLojda: Bug: when `--entities` is omitted, the fallback is `syncEntityChoices`, which still contains `"webSpotlight"`. `normalizeWebSpotlightAlias` then unconditionally logs `--entities webSpotlight is deprecated; use livePreview instead.` on every default-entities diff run, even though the user never typed the flag. `run`/`snapshot` aren't affected because `--entities` is `demandOption` there. Easiest fix: default to an alias-free list (e.g. `syncEntityChoices.filter(e => e !== legacyWebSpotlightAlias)`) before passing it through the normalizer.

src/modules/sync/constants/entities.ts
```diff
@@ -19,17 +21,17 @@ export const syncEntityChoices = [
   "assetFolders",
   "spaces",
   "languages",
+  "livePreview",
   "webSpotlight",
```
JiriLojda: Since `--help` on all three commands renders `Allowed entities are: ${syncEntityChoices.join(", ")}`, this entry advertises `webSpotlight` as a peer of `livePreview` with no deprecation hint. Worth either dropping it from the user-facing choices list (and accepting the alias via custom validation), or splitting into a separate `userVisibleEntities` array used for the help string.

src/modules/sync/constants/entities.ts
```diff
@@ -38,15 +39,15 @@ export const syncEntityDependencies: Record<
...
+  livePreview: ["livePreview"],
 };
 
 export const entityToFilename = {
   contentTypes: contentTypesFileName,
   contentTypeSnippets: contentTypeSnippetsFileName,
   taxonomies: taxonomiesFileName,
   collections: collectionsFileName,
-  webSpotlight: webSpotlightFileName,
+  livePreview: livePreviewFileName,
```
JiriLojda: We have a backwards compatibility in the API that when user supplies webSpotlight entity we convert that to livePreview, but we don't have the same backwards compatibility for loading files in the validation or file loading. Therefore if user runs the new data-ops version with the same parameters and the same input folder (webSpotlight.json file present) it will fail with a message that livePreview.json file is missing. That a) doesn't make sense, b) is a breaking change. We can either go with a major version bump and do all the breaking changes we need and document them properly or implement a proper backwards compat even in this part by properly loading webSpotlight.json file if that entity is provided and only loading livePreview.json file if livePreview entity is provided not for webSpotlight entity.

JiriLojda: If going the backwards compat route, we should (in the readme) mention that the root_type of webSpotlight.json files will be ignored, because the file will be used as a livePreview config instead of the disabled webSpotlight.

JiriLojda (changes requested):
We must update the readme files. There is still webspotlight option and no live preview option. Ideally also mention that the webspotlight option is deprecated and live preview is the replacement in the readme with a link to our docs where this is mentioned.

JiriLojda (approved):

PR #185, opened 2026-07-15
redact credentials in error output at serialization sinks

IvanKiral:
### Motivation

Error output could include sensitive request-header values. When a request fails, several sinks serialized the raw axios/Kontent error to stderr via `JSON.stringify`, which pulls in the full request headers.

**What:** Mask sensitive header values (`authorization`, `api-key`) wherever an error is serialized, while leaving every error's shape and type exactly as before.

**Why this approach:** Deliberately kept minimal and **non-breaking**. We considered introducing a dedicated error type and converting errors at the module boundary, but every module function has a public counterpart, so changing what they throw would be a breaking change — that's left for the next major. This PR only masks values, so it ships as a patch. The stderr output shape is also unchanged (a function replacer masks values in place rather than reshaping the object), so behavior is identical apart from the masked values.

**How:**
- Added `redactingReplacer` in `utils/error.ts` — masks sensitive keys at any depth — and applied it to the four bare `JSON.stringify(e)` sinks (`diff.ts`, `migrationUtils.ts`, `assetFolders.ts`, `errUtils.ts`).
- `simplifyAxiosErrors` now masks the sensitive request-header values in place (keeps the keys).
- Left the `JSON.stringify(e, Object.getOwnPropertyNames(e))` allowlist sinks untouched — they already omit nested header values, and changing them would alter output shape.

### Checklist

- [x] Code follows coding conventions held in this repo
- [x] Automated tests have been added
- [x] Tests are passing
- [ ] Docs have been updated (if applicable) — n/a
- [x] Temporary settings have been reverted to defaults

### How to test

Run any command against an environment with an invalid API key (e.g. `environment backup` or `migrations run`) and confirm the error output shows `"authorization": "***redacted***"`. Covered by `tests/unit/utils/error.test.ts`.

### DIFF
--- a/src/modules/migrations/utils/errUtils.ts
+++ b/src/modules/migrations/utils/errUtils.ts
@@ -1,4 +1,5 @@
 import { type LogOptions, logError } from "../../../log.js";
+import { redactingReplacer } from "../../../utils/error.js";
 
 type Err = { err: unknown };
 
@@ -8,7 +9,7 @@ export const handleErr = <T>(entity: WithErr<T>, logOptions: LogOptions, message
   if ("err" in entity) {
     logError(
       logOptions,
-      `${message ?? ""}${entity.err instanceof Error ? entity.err.message : JSON.stringify(entity.err)}`,
+      `${message ?? ""}${entity.err instanceof Error ? entity.err.message : JSON.stringify(entity.err, redactingReplacer)}`,
     );
 
     throw new Error(
--- a/src/utils/error.ts
+++ b/src/utils/error.ts
@@ -36,7 +36,28 @@ export const simplifyAxiosErrors = (error: unknown): unknown =>
         method: error.config?.method,
         responseBody: error.response?.config.data,
         requestBody: error.config?.data,
-        requestHeaders: error.config?.headers.normalize(true).toJSON(),
+        requestHeaders: redactSecrets(error.config?.headers.normalize(true).toJSON()),
         code: error.code,
       }
     : error;
+
+/**
+ * A `JSON.stringify` replacer that masks credential values at any depth. Pass it to every sink that
+ * serializes a raw error, whose nested request headers would otherwise expose the bearer token.
+ */
+export const redactingReplacer = (key: string, value: unknown): unknown =>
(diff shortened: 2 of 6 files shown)


src/modules/migrations/utils/errUtils.ts
```diff
@@ -8,7 +9,7 @@ export const handleErr = <T>(entity: WithErr<T>, logOptions: LogOptions, message
   if ("err" in entity) {
     logError(
       logOptions,
-      `${message ?? ""}${entity.err instanceof Error ? entity.err.message : JSON.stringify(entity.err)}`,
+      `${message ?? ""}${entity.err instanceof Error ? entity.err.message : JSON.stringify(entity.err, redactingReplacer)}`,
```
JiriLojda, line 12: Why don't you use the redactingReplacer also in the thrown error bellow? It could leak too.

src/utils/error.ts
```diff
@@ -36,7 +36,28 @@ export const simplifyAxiosErrors = (error: unknown): unknown =>
...
+const redactedValue = "***redacted***";
+
+const secretKeyPattern = /authorization|api[-_]?key/i;
+
+const redactSecrets = (headers: unknown): unknown =>
+  headers && typeof headers === "object"
+    ? Object.fromEntries(
+        Object.entries(headers).map(([key, value]) => [
+          key,
+          secretKeyPattern.test(key) ? redactedValue : value,
```
JiriLojda, line 60: Can use `redactingReplacer` here.

JiriLojda (changes requested):
The secret redaction in this PR is reliant on the assumption that the key is only in the Authorization header and that all sites that throw an uncaught error or log anything will diligently call the redact function. That is not very robust. What we can do instead:
Use the keys provided to the createClient functions (MAPI, DAPI keys) to be able to search those keys in any text (not just structured headers). With that, we can globally redact in all logs if we redact in the logInternal function (will need to propagate the keys there though) and can register the global [fail](https://yargs.js.org/docs/#api-reference-failfn-boolean) function to redact it in all thrown errors globally as well. What do you think?

JiriLojda:
Done in core sdk instead.

PR #186, opened 2026-07-17
SE-330 White-list allowed kontentUrl parameters

JiriLojda:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/src/utils/error.ts
+++ b/src/utils/error.ts
@@ -1,5 +1,8 @@
 import { SharedModels } from "@kontent-ai/management-sdk";
 import { isAxiosError } from "axios";
+import chalk from "chalk";
+
+import { InvalidKontentUrlError } from "./kontentUrl.js";
 
 export const throwError = (msg: string) => {
   throw new Error(msg);
@@ -24,9 +27,19 @@ export const handleKontentErrors =
     throw simplifyAxiosErrors(error);
   };
 
-export const simplifyErrors = handleKontentErrors((err) => {
+const rethrowKontentErrors = handleKontentErrors((err) => {
   throw err as unknown;
 });
+
+export const simplifyErrors = (error: unknown) => {
+  if (error instanceof InvalidKontentUrlError) {
+    // eslint-disable-next-line no-restricted-syntax
+    console.error(`${chalk.red("Error:")} ${error.message}\n`);
+    process.exit(1);
+  }
+
+  return rethrowKontentErrors(error);
+};
 
 export const simplifyAxiosErrors = (error: unknown): unknown =>
   isAxiosError(error)
--- a/src/utils/error.ts
+++ b/src/utils/error.ts
@@ -1,7 +1,7 @@
 import { SharedModels } from "@kontent-ai/management-sdk";
 import { isAxiosError } from "axios";
-import chalk from "chalk";
 
+import { type LogOptions, logError } from "../log.js";
 import { InvalidKontentUrlError } from "./kontentUrl.js";
 
 export const throwError = (msg: string) => {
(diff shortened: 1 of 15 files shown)


src/utils/error.ts
```diff
@@ -24,10 +27,20 @@ export const handleKontentErrors =
...
   };
 
-export const simplifyErrors = handleKontentErrors((err) => {
+const rethrowKontentErrors = handleKontentErrors((err) => {
   throw err as unknown;
 });
 
+export const simplifyErrors = (error: unknown) => {
+  if (error instanceof InvalidKontentUrlError) {
+    // eslint-disable-next-line no-restricted-syntax
```
IvanKiral: i think we have this restricted for a reason? :D should we use logError for this?

src/utils/error.ts
```diff
@@ -24,10 +27,20 @@ export const handleKontentErrors =
...
+    console.error(`${chalk.red("Error:")} ${error.message}\n`);
+    process.exit(1);
```
IvanKiral, line 37: do we want to exit here? I usually try to exist on the edges like in the cli or public commands

src/utils/kontentUrl.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+const parseUrl = (rawUrl: string): URL => {
+  const withoutScheme = rawUrl.replace(/^https?:\/\//, "");
+
+  try {
+    return new URL(`https://${withoutScheme}`);
+  } catch {
+    throw new InvalidKontentUrlError(
+      `Invalid kontentUrl "${rawUrl}": the value could not be parsed as a URL. The kontentUrl must be within an allowed domain: ${allowedKontentDomains.join(", ")}.`,
```
IvanKiral: why this throw depends on allowedKontentDomains when it is not in the try body?

tests/unit/utils/kontentUrl.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  ] as const)("throws InvalidKontentUrlError for disallowed input $input", ({ input }) => {
+    expect(() => validateKontentUrl(input)).toThrow(InvalidKontentUrlError);
+  });
+
+  it.each([
+    { input: "127.0.0.1" },
+    { input: "https://127.0.0.1" },
+    { input: "169.254.169.254" }, // cloud metadata endpoint
+    { input: "192.168.0.1" },
+    { input: "10.0.0.1" },
```
IvanKiral, line 39: should i be able to use data-ops against localhots?

IvanKiral (changes requested):

src/utils/error.ts
JiriLojda: Good point — switched to `logError` (adf896e). `simplifyErrors` now takes `LogOptions` and the call sites pass `args`, so the message respects `--logLevel` and the `no-restricted-syntax` disable is gone.

src/utils/error.ts
JiriLojda: Kept the `process.exit` but the reporting now goes through `logError` (adf896e). `simplifyErrors` is only ever used as the `.catch()` at the CLI command edge (never from `public.ts`/`modules`), so this *is* the edge. It ends up here because `createClient` (which runs the validation) is called outside each command's own try/catch. Happy to move it further out if you'd prefer.

src/utils/kontentUrl.ts
JiriLojda: Fixed in 484e68a — the `try` only attempts URL parsing, so the allowlist wording didn't belong here. Trimmed it to just the parse error; the allowed-domains message stays on the `isAllowed` check in `validateKontentUrl`.

tests/unit/utils/kontentUrl.test.ts
JiriLojda: By design — blocking localhost/loopback/private IPs (and the cloud-metadata endpoint) is the point of SE-330: it stops the API key from being sent to an attacker-controlled host. Leaving it blocked.

JiriLojda: I don't think we would test very often against locally-run backend. For those cases it should be comparatively easy to just add localhost to the constant temporarily.

IvanKiral (approved):

PR #187, opened 2026-07-30
Update deps

JiriLojda:
### Motivation

### Checklist

- [ ] Tests are passing

### How to test


src/modules/migrations/models/migration.ts
```diff
@@ -15,8 +15,10 @@ export type MigrationModule = Readonly<{
 
 const migrationModuleSchema: z.Schema<MigrationModule> = z.object({
   order: z.union([z.number(), z.coerce.date()]),
-  run: z.function().args(z.custom<ManagementClient>()).returns(z.promise(z.void())),
-  rollback: z.function().args(z.custom<ManagementClient>()).returns(z.promise(z.void())).optional(),
+  run: z.custom<(apiClient: ManagementClient) => Promise<void>>((v) => typeof v === "function"),
```
IvanKiral, line 18: did you check this? not sure whether we have test for this

eslint.config.js
```diff
@@ -0,0 +1,64 @@
+import js from "@eslint/js";
+import tsPlugin from "@typescript-eslint/eslint-plugin";
+import tsParser from "@typescript-eslint/parser";
```
IvanKiral, line 4: kontent-ai/eslint-config seems not to be used

src/utils/typeguards.ts
```diff
@@ -54,10 +54,9 @@ const baseTypeSchema = z.object({
   codename: z.string().optional(),
 });
 
-const snippetSchema: z.ZodType<ContentTypeSnippetModels.IAddContentTypeSnippetData> =
-  baseTypeSchema;
+const snippetSchema = baseTypeSchema;
 
-const contentTypeSchema: z.ZodType<ContentTypeModels.IAddContentTypeData> = baseTypeSchema.extend({
```
IvanKiral, line 60: why remove this type? it used to be there for matching sdk contracts? is there areson for this?

src/modules/sync/syncSnapshot.ts
```diff
@@ -74,6 +74,8 @@ const logOnError = <T>(errorMessage: string, action: () => Promise<T>): Promise<
   try {
     return action();
```
IvanKiral: should we add await action here? becaus now it just returns apromise so anything inside the actual action won't be catched?

IvanKiral (changes requested):

eslint.config.js
JiriLojda: I added it and fixed all the errors that appeared after that :D

src/modules/sync/syncSnapshot.ts
JiriLojda: I think it is added now.

src/modules/migrations/models/migration.ts
JiriLojda: Based on the docs it should be. https://zod.dev/api?id=custom#custom But we don't have any tests on that. Claude tested it manually though and it worked.

src/utils/typeguards.ts
JiriLojda: I had to remove it, because we don't properly check the elements property and so they are typed unknown[] which is not assignable there. So we can either make the check proper or adjust the type, but I thought that it is ok to remove it now.

src/modules/migrations/models/migration.ts
JiriLojda: But I added integration tests for migrations to be sure :)

src/log.ts
```diff
@@ -50,6 +50,9 @@ const optionsToLogLevel = (options: LogOptions): LogLevel => {
     return "verbose";
   }
 
+  // `logLevel` is a raw string; an empty string should also fall back to the default,
```
IvanKiral, line 53: should this be a string though? in commands we have demand options, therefore,  logLevel should be always none, standar or verbose. 

The situation is worse from api functions, in which probably every string is allowed, though adding a correct type could fix that?

src/modules/migrations/models/migration.ts
IvanKiral: can we add 1-2 tests for status plugins as well?

IvanKiral (approved):

IvanKiral (approved):

PR #188, opened 2026-08-14
Add validation command

JiriLojda:
### Motivation

Provide entity zod schemas in the public interface to enable user-side validation of sync files before the actual sync so users can do validation without syncing anything.


package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "build": "tsc",
+    "prepare": "pnpm run build",
```
IvanKiral: should we still keep this?

IvanKiral (changes requested):

package.json
JiriLojda: Removed

IvanKiral (approved):

